PR #1, opened 2017-11-23
Fix cfi-icall breakage

lattera:
Signed-off-by:	Shawn Webb <shawn.webb@hardenedbsd.org>

### DIFF
--- a/command.c
+++ b/command.c
@@ -799,7 +799,7 @@ getcc_end_command()
  * (characters previously given to ungetcc or ungetsc).
  */
 	static LWCHAR
-getccu()
+getccu(VOID_PARAM)
 {
 	LWCHAR c;
 	if (ungot == NULL)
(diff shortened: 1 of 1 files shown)


gwsw:
Thanks for the patch Shawn.  One question: I understand the change to add VOID_PARAM to the getccu prototype, but was there a reason you changed the four calls through the function pointers from the (*gr_getc)() style to gr_getc()?  AFAIK they are equivalent but the former is the style used elsewhere in less so I'd prefer to keep it consistent.

lattera:
I can remove that. It was a stylistic thing that was bugging me while I was debugging the issue. Consider it leftover cruft.

Would you like me to update this PR to undo the stylistic changes? I'd be more than happy to do so.

gwsw:
Yes, if you could trim it down to the necessary change, I'd appreciate it.

lattera:
Done. :)

PR #6, opened 2019-05-30
Unlock scrolling when input fd is closed.

aleclearmind:
As per our discussion, here's a PR so we can discuss more easily.
If you can point me to an easy way to test this change on other platforms, I'd be happy to help.

Please also consider the alternative option I suggested: when in follow mode and an arrow key is pressed, `less` could simply leave the follow mode (right now I can achieve this by pressing `Ctrl + Z` and then `fg`). No need for `select`/`poll` in that case.

### DIFF
--- a/ch.c
+++ b/ch.c
@@ -19,6 +19,8 @@ extern dev_t curr_dev;
 extern ino_t curr_ino;
 #endif
 
+#include <poll.h>
+
 typedef POSITION BLOCKNUM;
 
 public int ignore_eoi;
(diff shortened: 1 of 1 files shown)


gwsw:
I don't know of a way to implement your last suggestion without select/poll or something similar. While less is sitting in a read() waiting for file input, pressing keys won't interrupt the read, so it has no chance to see the keypress until the read completes. ctrl-C or ctrl-Z interrupts the read, but ordinary input chars do not.

gwsw:
This is now implemented in df1494f4825857bb3db1b025f16ecefe4265d6dc.

PR #17, opened 2019-07-16
Use ANSI prototypes for static functions.

brooksdavis:
Some ABIs require prototypes in order to call functions correctly.

This completes the work begun in b1f1dfba78310a515e90cc007b30a10ba9d798d0.


brooksdavis:
Thanks @gwsw.

PR #18, opened 2019-07-21
Correct description of -S

george-b:
Lines are not truncated with -S, diff here matches change applied in OpenBSD by Ingo Schwarze.

https://marc.info/?l=openbsd-tech&m=156362993300918&w=2

### DIFF
--- a/less.nro.VER
+++ b/less.nro.VER
@@ -801,10 +801,8 @@ This is useful when viewing
 .I nroff
 output.
 .IP "\-S or \-\-chop-long-lines"
-Causes lines longer than the screen width to be
-chopped (truncated) rather than wrapped.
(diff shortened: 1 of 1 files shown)


gwsw:
I don't think the proposed wording completely solves the problem. If someone is pedantically interpreting "truncated" as permanently removing the undisplayed portion of the file, they might equally well object that "Display only the beginning of lines that exceed the screen width." is correct only immediately on entry; after you press an arrow key you are no longer seeing the beginning of the line. (Also the rscroll feature added in less-530 makes it more clear that there is undisplayed data.) I have changed the wording to try to be clearer.

george-b:
Thanks for making the update.

PR #29, opened 2019-10-17
Default to Windows virtual terminal processing

lzybkr:
Bypass the legacy Windows virtual terminal processing if the Windows
console supports virtual terminal processing.

This fixes issues exposed by using https://github.com/dandavison/delta as the pager for git on Windows.

### DIFF
--- a/output.c
+++ b/output.c
@@ -35,7 +35,7 @@ extern int so_fg_color, so_bg_color;
 extern int bl_fg_color, bl_bg_color;
 extern int sgr_mode;
 #if MSDOS_COMPILER==WIN32C
-extern int have_ul;
+extern int vt_enabled;
 #endif
 #endif
 
@@ -115,7 +115,7 @@ flush(VOID_PARAM)
 	if (is_tty && any_display)
 	{
 		*ob = '\0';
-		if (ctldisp != OPT_ONPLUS)
+		if (ctldisp != OPT_ONPLUS || vt_enabled)
 			WIN32textout(obuf, ob - obuf);
 		else
 		{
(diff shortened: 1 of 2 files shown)


lzybkr:
@gwsw - Have you had a chance to review this PR yet?

gwsw:
A reviewer had this comment:

> There could be an issue that a program calling "less -X" will then unexpectedly have terminal processing enabled (CMD is unaffected, as it restores the mode itself).  I think the no_init init output_mode needs to be preserved, so a no_init deinit can restore it.

lzybkr:
@gwsw - Sorry for the delay and I agree the console mode should be restored. I made the change to restore the initial console mode on both paths for future maintainability and in case the !no_init path call to SetConsoleMode is applied to the same handle as the no_init path.

adoxa:
Another issue, that I'm amazed hasn't been mentioned, but find it hard to believe is just me, is that with terminal emulation enabled wrap is different. This causes the prompt line to occasionally shift up. Initial testing suggests a simple `autowrap = 0; ignaw = 1;` will fix it, so you'd probably now want a new function to set terminal emulation and those flags as well.

adoxa:
Two other issues have cropped up: since you wrote this there's been a fix for `-F`, which changes how (de)init is called for a single screen; and it no longer uses the `-D` colours (added test for `sgr_mode`, too). (Here's)[https://github.com/adoxa/less/commit/4c1a31991487972366fd882192d78c1a7714cfb6] my approach.

gwsw:
Merged with some changes from Jason Hood, in 23dac1c8a0e128329458f952353ae3c7c60c6e19.

lzybkr:
Thanks. I forgot about this PR in part I think because I was no longer seeing the issue for reasons I can't explain. I'm pretty sure my private build of less was overwritten by an update to git, yet the original issue I was fixing no long reproduced.

Regardless, the change seems right, so I'm glad it was merged.

PR #35, opened 2019-11-26
Prevent double free of namelogfile

mmuzila:
In certain circumstances [1] less crashes because it tries to
double-free memory allocated at 'namelogfile'. Prevent this by setting
'namelogfile' to NULL after each free.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1770901

### DIFF
--- a/edit.c
+++ b/edit.c
@@ -847,7 +847,6 @@ loop:
 		/*
 		 * Don't do anything.
 		 */
-		free(filename);
 		return;
 	case 'q':
 		quit(QUIT_OK);


gwsw:
Would a simpler solution be to simply remove the call to free() in use_logfile? I don't think it belongs there. It's only executed when the user presses 'D'; none of the other paths call free(), nor do they save the parameter anywhere. So I don't see why it should be calling free() in just this one circumstance. The parameter is always the namelogfile global anyway, so it doesn't seem like it could get leaked.

mmuzila:
Yes, you're right. It would be simpler to remove call to `free()` in `use_logfile`.  I'm going to change the patch.

I got confused by the `QUERY` case in `opt_o`, but I can (now) see that setting a new log file is not possible if a log file is already in use. (If it was possible, invalid file name of log file could be displayed on `_o` command).

PR #45, opened 2020-01-24
Make it possible to build with Makefile.wnm from a Git checkout

SpecLad:


gwsw:
Added README instructions for building from a git checkout in ada582c0360a639414eb37749e392e6858263743.

PR #51, opened 2020-03-04
Quit less if the file fits in less than X lines

tqdv:
This was a feature I wanted, so I implemented it.
I don't really know how less works, so I based it off the quit-if-one-screen code.

This is essentially a pull request to inform you that I patched in a feature, and you can merge if you want.


yabeenico:
I really want this function.
Why don't you reopen this PR (if possible).

Issue #54, opened 2020-04-09
Feature Request: Add support for hyperlinks with `-R`


tomprince:
There is a standard for displaying hyperlinks in terminals [here](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda). It would be nice if less passed through those escapes when `-R` is passed, like it does colors.

(Originally reported as 308 on the [bug list](http://www.greenwoodsoftware.com/less/bugs.html))

There is a patch [here](https://bug779734.bugzilla-attachments.gnome.org/attachment.cgi?id=349890) (from [here](https://bugzilla.gnome.org/show_bug.cgi?id=779734)).

poettering:
Note that bigger projects including gcc and systemd nowadays generate hyperlinks like that in their output. It would be excellent if less would handle this correctly!

(I am systemd maintainer)

dandavison:
In case it helps the process of getting this feature into master, I have applied the patch (with minor modifications to fix the conflicts) on master, confirmed that hyperlinks are working, and opened a PR: https://github.com/gwsw/less/pull/75.

(My motivation for this is that I'm the author of an application that relies on `less` for paging and would like to use hyperlinks: https://github.com/dandavison/delta/pull/261.)

torarnv:
If adding to what `-R` does is seen as too risky for backwards compatibility, perhaps a new environment variable similar to `LESSANSIENDCHARS`and `LESSANSIMIDCHARS` could be added that controls which control codes to pass through, defaulting to the existing  ANSI color escape sequences?

gwsw:
This is implemented in 3bb781ea44461dea611c4b1c5b414f97fc396cbe. My tests with self-constructed files seem to work fine, but could someone who regularly uses files containing such hyperlinks confirm that the behavior is as expected?

dandavison:
Thanks very much for working on this @gwsw. I have built 3bb781ea44461dea611c4b1c5b414f97fc396cbe locally and can confirm that it is working as expected for me.

(The way I tested was a bit idiosyncratic, but for the record I used [delta](https://github.com/dandavison/delta)  with the setting `hyperlinks = true` and issued `git show`. With the version of less that I had installed already, the hyperlink was displayed correctly only if I used the `-r` flag, but not with `-R`. But with the version of less built from  3bb781ea44461dea611c4b1c5b414f97fc396cbe the hyperlink was displayed correctly when I used `less -R` and also `less -r`. This was on MacOS under iTerm2.).

[...]

PR #75, opened 2020-07-22
Apply OSC 8 hyperlinks patch

dandavison:
Ref https://github.com/gwsw/less/issues/54

In case it is useful, this PR applies the OSC hyperlinks patch from https://bug779734.bugzilla-attachments.gnome.org/attachment.cgi?id=349890 on top of master. (Some minor edits were required to make the patch apply.)

I am using less built from this commit, and can confirm that hyperlinks are passed on correctly and that I have not observed any problems with less while using it so far. However, I am not familiar with the less code base and I am not sure how to test thoroughly. Please feel free to close this PR if it is just creating noise.

### DIFF
--- a/charset.c
+++ b/charset.c
@@ -426,6 +426,7 @@ control_char(c)
 	LWCHAR c;
 {
 	c &= 0377;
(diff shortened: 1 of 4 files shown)


gwsw:
Reimplemented as of 65d73a2b54ddb390993738d9c6462af9a9661989

PR #82, opened 2020-10-18
Add wrap-around search modifier

numirias:
This PR implements wrap-around search for the current file.

With the added `^W` search modifier in use, a forward search doesn't end at EOF, but continues from the beginning of the current file up until the searches' starting point. A backward search does the reverse.

Rationale: Wrap-around is a common search feature that has been [sought for in less in the past](https://unix.stackexchange.com/questions/255862/search-in-less-with-wrap-around-at-eof). It allows the user to use the same key (`n`/`N`) to infinitely loop through the search results, without having to consider that they may have already gone past a match. The feature is limited to the current file because the added complexity seems reasonable. It looks to me like wrap-around across multiple files would come with some gotchas and be less straightforward.

### DIFF
--- a/command.c
+++ b/command.c
@@ -168,6 +168,8 @@ mca_search(VOID_PARAM)
 		cmd_putstr("Keep-pos ");
 	if (search_type & SRCH_NO_REGEX)
 		cmd_putstr("Regex-off ");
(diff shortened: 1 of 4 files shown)


agguser:
Should also have a notification message in the status bar when wrap-around occurs, like vim.

PR #87, opened 2020-10-31
Multiple &-filters can now be in effect at the same time.

folkol:
A common use case for me is that I want to filter some "spammy lines" out while reading log files, but more often than not next category of spammy lines emerges just after the first filter is applied. In these cases I might try to update the filter pattern to cover both line patterns, which might be non-trivial — for example if the individual patterns is using a mix of ^R and ^N. (In reality, I tend to just quit less and pre-filter the log file instead). The change in this pull request saves me from doing most of those context switches — and I figured that someone else might have the same problem.

### DIFF
--- a/command.c
+++ b/command.c
@@ -780,8 +780,15 @@ prompt(VOID_PARAM)
 	clear_cmd();
 	forw_prompt = 0;
 	p = pr_string();
-	if (is_filtering())
-		putstr("& ");
+	switch (is_filtering()) {
+		case 0:
+			break;
+		case 1:
+			putstr("& ");
+			break;
+		default:
+			putstr("&* ");
+	}
 	if (p == NULL || *p == '\0')
 		putchr(':');
 	else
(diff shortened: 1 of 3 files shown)


folkol:
I just found a bug (I can no longer toggle search highlights with ESC-u), I will close this pull request and make a new one that isn't as premature. Sorry for the line noise.

folkol:
While investigating the bug above, I found that it was still present if I removed all my changes — and it is present in the less version from Homebrew on Mac (version 551).

gwsw:
Thanks for this. I have merged most of your code, but allocate the filter structs dynamically as you suggested in a comment. For now, I have also opted not to change the prompt when there is more than one filter, as I don't really see the need to distinguish between 1 and 2 filters, but not 2 and 3 etc. Fixed in 6a070fc53799fb86e0fe38806d52b7ab72bd3aa4.

PR #92, opened 2020-11-13
Fix configure for Xcode 12

### LINKED ISSUE #91: ncurses linkage fails with Xcode 12
`less` fails to build on macOS 10.15 as Xcode 12 no longer allows implicit function declarations. This prevents `configure` from successfully linking with `ncurses`.

This is fixed in #92 by adding `#include <termcap.h>`

Output of `./configure`:

```
checking for gcc... clang
checking whether the C compiler works... yes
checking for C compiler default output file name... a.out
checking for suffix of executables... 
checking whether we are cross compiling... no
checking for suffix of object files... o
checking whether we are using the GNU C compiler... yes
checking whether clang accepts -g... yes
checking for clang option to accept ISO C89... none needed
checking for library containing strerror... none required
checking how to run the C preprocessor... clang -E
checking for grep that handles long lines and -e... /usr/bin/grep
checking for egrep... /usr/bin/grep -E
checking whether clang needs -traditional... no
checking for a BSD-compatible install... /usr/bin/install -c
checking for special C compiler options needed for large files... no
checking for _FILE_OFFSET_BITS value needed for large files... no
checking for tgoto in -ltinfo... no
checking for tgoto in -ltinfow... no
checking for initscr in -lxcurses... no
checking for initscr in -lncursesw... no
checking for initscr in -lncurses... yes
checking for initscr in -lcurses... yes
checking for tgetent in -ltermcap... yes
checking for tgetent in -ltermlib... no
checking for library containing regcmp... no
checking for working terminal libraries... Cannot find terminal libraries - configure failed
```

SeekingMeaning:
Fixes #91

This adds `#include <termcap.h>` so that `tgetent`/`tgetflag`/`tgetnum`/`tgetstr` can be found.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -110,7 +110,7 @@ if test "x$TERMLIBS" = x; then
     TERMLIBS="-lncurses"
     SAVE_LIBS=$LIBS
     LIBS="$LIBS $TERMLIBS"
-    AC_TRY_LINK(, [tgetent(0,0); tgetflag(0); tgetnum(0); tgetstr(0,0);],
+    AC_TRY_LINK([#include <termcap.h>], [tgetent(0,0); tgetflag(0); tgetnum(0); tgetstr(0,0);],
       [termok=yes], [termok=no])
     LIBS=$SAVE_LIBS
     if test $termok = no; then TERMLIBS=""; fi


sharris40:
The include should probably be `term.h` instead of `termcap.h`, since that’s what’s in the man pages for `ncurses` (and X/Open Curses).

If you want to fix up the other termcap checks to be C99-compatible, I believe `term.h` is pretty universal, the only exception I found being an obsolete GNU termcap library supporting only `termcap.h` that links as either `-ltermcap` or `-ltermlib` depending on the system. NetBSD’s man pages also specify `termcap.h` (and will link as both `-ltermcap` and `-ltermlib`), although `term.h` is supported. Every other manual I found uses `term.h`.

SeekingMeaning:
I just checked MacPorts, and their patch uses `curses.h` and `term.h` : https://github.com/macports/macports-ports/blob/master/textproc/less/files/patch-configure.diff

sharris40:
Eh, thinking about it, it’s probably better left as is. Termcap has already been deprecated for decades, and this codebase already uses `termcap.h`. A better idea would be to use `terminfo` functions instead of `termcap` if available, but that’s beyond the scope of this pull request.

gwsw:
I wonder if some or all of the other AC_TRY_LINKS for term libraries should add the #include <termcap.h>. It seems like a good idea in case other systems adopt stricter compilers, but there's a potential that it could fail on currently working systems which happen to not have termcap.h, so I'm reluctant to change them.

PR #121, opened 2021-02-05
Treat byte order mark (BOM) as binary

ZekeLu:
If the BOM is  added to/removed  from a file, `git diff` will report it like this (with the excellent `less` as the pager):

```diff
diff --git a/copyright b/copyright
index a3cfb63..4602453 100644
--- a/copyright
+++ b/copyright
@@ -1,4 +1,4 @@
-/*
+﻿/*
  * Copyright (C) 1984-2021  Mark Nudelman
  *
  * You may distribute under the terms of either the GNU General Public
```

It's impossible to find out what's changed from the output.

I have checked the change history, and found that the behavior is changed  by the commit 07c14bb6bc1019195f896e53789819d8dc9bf584. Before this commit, it will produce the following output:

```diff
diff --git a/copyright b/copyright
index a3cfb63..4602453 100644
--- a/copyright
+++ b/copyright
@@ -1,4 +1,4 @@
-/*
+<U+FEFF>/*
  * Copyright (C) 1984-2021  Mark Nudelman
  *
  * You may distribute under the terms of either the GNU General Public
```

According to that commit, I have tried the `-U` option. But it will transform other formatting chars too. Here is the output with the `-U` option (I added a `tab` before `see  the README file` and the `tab` is displayed as `^I`):

```diff
diff --git a/copyright b/copyright
index a3cfb63..d586b01 100644
--- a/copyright
+++ b/copyright
@@ -1,8 +1,8 @@
-/*
+<U+FEFF>/*
  * Copyright (C) 1984-2021  Mark Nudelman
  *
  * You may distribute under the terms of either the GNU General Public
  * License or the Less License, as specified in the README file.
  *
- * For more information, see the README file.
+ * For more information,^Isee the README file.
  */
```

This pull request forces it to treat the BOM as a binary char, so that it will be displayed as `<U+FEFF>`.


gwsw:
I'm not sure about this. I would think someone doing normal file viewing would not want to see a U+ code for the BOM, which might appear in many of their files depending on their tools. On the other hand, someone viewing a git diff and being puzzled by no apparent difference could toggle -U and see the BOM change, and also see the exact sequence of spaces/tabs that appear as spaces in the output, etc. In other words, I think the fact that -U shows exactly what's in the file is a feature, not a bug. The intention is that the default mode (lack of -U) shows a "readable" version of the file, while -U shows the details.

ZekeLu:
> The intention is that the default mode (lack of -U) shows a "readable" version of the file, while -U shows the details.

Thanks for the comment! That's reasonable.

Please take the following points into account  when making the decision:

1. No matter the BOM is encouraged or discouraged, it's important to make it visible for a git user.
2. Since the `-U` option will make other charecters visible too, it's not a good choice to configure git to enable the `-U` option always.
3. This behaviour is not documented and it's not easy to figure out why the `<U+FEFF>` sequence magically disappeared (for a user that is used to it).

gwsw:
Well, regarding points 1 and 2, I'm not sure I see why it's important that a git user see the BOM, but it's not important that they see other nonvisible changes, such as tabs that have been changed into spaces or vice versa. That seems like a similar sort of issue to me, and is similarly solved by enabling -U.
Regarding point 3, I have added a line to the man page to be more explicit about how -U affects formatting characters.

ZekeLu:
Okay, it looks more like a style problem. I'm going to close the PR. Thank you very much!

Issue #140, opened 2021-03-26
`grep --color=always ... | less -R` garbled since 580


vain:
How to reproduce:

```
$ echo hello world >foo
$ grep --color=always . foo | less -R
(END)
```

I would expect it to print that one matching line (colored):
```
$ echo hello world >foo
$ grep --color=always . foo | less -R
hello world
(END)
```

It’s because `grep` uses lots of “Erase in Line” escape sequences:

https://git.savannah.gnu.org/cgit/grep.git/tree/src/grep.c?id=54aac0789909618122d7125854c443f26a94f610#n317

I am not sure how “Erase in Line” is *supposed* to be interpreted. My VTE terminal and XTerm only erase characters that were already on screen:

```
$ printf 'hello\ra\n'
aello
$ printf 'hello\r\e[Ka\n'
a
```

Piping the second command through `less` erases the `a` as well and I guess that’s not correct?

(I also see some more corruption like blinking output or misplaced cursors, but I have no idea why that happens, so I can’t provide a meaningful but report for that.)

Let me know if there’s anything else I can do to help.

Thanks!

– edit: Forgot to mention it, this is on Arch Linux with less 580, grep 3.6, XTerm 366, and vte3 0.64.0.

heftig:
cc: @keszybz @tomprince @poettering

Arch upgraded less to support systemd's hyperlinks (#54). Is Fedora now also suffering from this bug?

heftig:
Bisection led to the commits implementing hyperlink support.

The first bad commit could be any of: 0f810ef16781bf0f59690be63af876bddabf68bf 3087a2c66085b3e1831dd976ed72671efaf7a269

gwsw:
This is mostly fixed in 792886416198c61c55d2e7768d5280efecb93888. Thanks to @heftig for the fix. 
Horizontal scrolling is still not working with colored grep output. This is apparently due to the (entirely unnecessary) line erase sequences that grep is adding to the output for some reason. As documented, less -R is not guaranteed to work correctly in the presence of escape sequences other than SGR coloring sequences. If the line erase sequences are removed it works correctly:
```
echo hello | grep --color=always . | sed 's/\o33\[K//g'  | less -R
```

I will investigate whether there is some easy workaround to ignore these line erase sequences, but it's rather annoying that grep is putting this junk into its output.

heftig:
[...]

PR #142, opened 2021-03-28
Fix ANSI skipping

heftig:
- Add a missing `ansi_done` to `bin_file` to free `pansi`.
 - Fix the inverted loop condition in `skip_ansi` so it actually skips.
 - Fix a condition in `step_ansi`: `pansi->hindex == 0` was never true.
 - Fix the sequence removal in `store_ansi` to reset the `line_ansi`.

Fixes https://github.com/gwsw/less/issues/140

### DIFF
--- a/filename.c
+++ b/filename.c
@@ -481,8 +481,10 @@ bin_file(f)
 			LWCHAR c = step_char(&p, +1, edata);
 			struct ansi_state *pansi;
 			if (ctldisp == OPT_ONPLUS && (pansi = ansi_start(c)) != NULL)
+			{
 				skip_ansi(pansi, &p, edata);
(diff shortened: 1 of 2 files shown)


heftig:
Hmm, horizontal scrolling is still broken.

heftig:
Why are the `ansi_state` structs heap-allocated at all? Seems unnecessary costly given that the state can be static or stack-allocated.

The code seems to be confused about whether the first `ansi_step` needs to be called with the escape character that started the sequence or the following character. It would be cleaner if it consistently started stepping with the first character after the escape.

Anyway, I don't understand the scrolling; I'm giving up on this.

PR #180, opened 2021-06-11
Add --repeat-search-skip-screen option.

rosche:
The --repeat-search-skip-screen option causes repeated search (N/n) to start on
the next page.  I don't think that's possible with --search-skip-screen.

Normally I don't need to go through the matches one-by-one, I just want to see
where they are.  Because I've already seen all the matches on the screen I
don't need those pointed out again.  Skipping them saves a lot of time without
losing information.

### DIFF
--- a/less.hlp
+++ b/less.hlp
@@ -219,6 +219,8 @@
                   Don't send termcap keypad init/deinit strings.
                 --no-histdups
                   Remove duplicates from command history.
+                --repeat-search-skip-screen
+                  Repeated search (n/N) skips current screen.
                 --rscroll=C
                   Set the character used to mark truncated lines.
                 --save-marks
--- a/less.nro.VER
+++ b/less.nro.VER
@@ -518,8 +518,9 @@ it may be necessary to quote the question mark, thus: "\-\e?".)
 By default, forward searches start at the top of the displayed screen
 and backwards searches start at the bottom of the displayed screen
 (except for repeated searches invoked by the n or N commands,
-which start after or before the "target" line respectively;
-see the \-j option for more about the target line).
+which by default start after or before the "target" line respectively;
+see the \-j option for more about the target line
+and \-\-repeat-search-skip-screen to modify this behavior).
 The \-a option causes forward searches to instead start at
 the bottom of the screen
 and backward searches to start at the top of the screen,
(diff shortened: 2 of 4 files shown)


gwsw:
Can you explain the difference between this and -a? Because "causes repeated search (N/n) to start on
the next page" seems to me to be exactly what -a does.

rosche:
For the the initial (/ or ?) search this is what -a does.  However when
you repeat the search with n less starts searching after the target line.  I
didn't see a way to modify this second behavior.

As it is now if I search for something and there are 20 of them highlighted
(highlit?) on the screen then to see further matches I have to either page
forward then hit n, or hit n 20 times.  With this patch I can see the 20 on the
screen and hit n once to see new matches.

I'll feel a fool if it was already possible to do what I'm describing (I tried
to figure out a way) , but oh well!

BTW I've been using less for 30 years and can hardly imagine what life would be
like if all we had was more.  So, so many thanks!

gwsw:
Hm, that doesn't seem to be the way -a acts for me. For example, in a 40 line window, if I view main.c in the less source with
```
less -a -N -j1 main.c
```
and then type ```/char``` to search for "char", it puts line 42 at the top of the screen and displays lines 42 through 80, with additional matches shown on lines 43, 47, 48, 53, 72 and 75. If I then type ```n```, it skips all the matches between lines 42 and 80 and puts line 101 at the top of the screen.  Am I misunderstanding something?

rosche:
Blargh.  I'd somehow ended up without -a in my $LESS then didn't properly test it with no $LESS.  Sorry about that!

rosche:
This problem has continued to happen, I wasn't explaining it correctly.  I wanted to dig deep to make extra extra sure I wasn't embarrassing myself again, but that might never happen.  So I take the risk.

I keep my target line in the middle of the screen.  If I search for a string which is in the top half of the screen but not the bottom half, less immediately scrolls to the next match off the screen.  It should highlight the upper match and stay on the page for the initial search.

E.g. the string "back" appears on line 21 of the man page on my screen, the target line is 23.  I "/back" and it scrolls down to line 53 (the text explaining the b command) and I never see the first match.

rosche:
I'll open a new pull request, I was subconsciously thinking a new comment would re-open this one.

rosche:
NM, I just saw "reopen and comment".  Sorry for missing that.

PR #185, opened 2021-07-02
Don't block on files that are not regular or block devices

4D617278:
Less currently freezes if a named pipe on Linux does not have any input. 

`$ mkfifo /tmp/test`
`$ less -f /tmp/test`

This will make less nonblocking when reading from the named pipe.
Tested on OpenBSD and Gentoo Linux.


4D617278:
Never mind, I don't have a need for this anymore

PR #191, opened 2021-08-15
Don't add a new line filter if the pattern is invalid

marko-turk:
less will crash if an invalid filter is added with & command.

To reproduce:
    - enter invalid pattern with &+
    - refresh screen (ctrl+L or similar)


gwsw:
Fixed in 74c598dd717de5e00db3d4995ee23d01f3510516.

PR #203, opened 2021-09-26
Fixed out of bound read with short lesskey file

d-vb:
Proof of concept:
Compiled with -fsanitize=address

echo AE0r | base64 -d > lesskey.txt
less -k lesskey.txt

Shoutout to [@c3h2_ctf](
https://twitter.com/c3h2_ctf
)


gwsw:
Fixed in 7318ae5ce310fe8a8784a8b0c80132099b11862c. A valid pre-v241 lesskey file could be 3 bytes long, although it's exceedingly unlikely that anyone is still using one.

PR #207, opened 2021-09-26
Fix memory leaks

stoeckmann:
If LESS or MORE environment variable is set, then a copy of the value
is created for scan_option call. Release memory after scan_option has
been performed.

Also handle rare cases in which a history file has duplicated mark
entries.

### DIFF
--- a/main.c
+++ b/main.c
@@ -140,7 +140,7 @@ main(argc, argv)
 
 	s = lgetenv(less_is_more ? "MORE" : "LESS");
 	if (s != NULL)
-		scan_option(save(s));
+		scan_option(s);
 
 #define isoptstring(s)  (((s)[0] == '-' || (s)[0] == '+') && (s)[1] != '\0')
 	while (argc > 0 && (isoptstring(*argv) || isoptpending()))
--- a/mark.c
+++ b/mark.c
@@ -59,6 +59,7 @@ cmark(m, ifile, pos, ln)
 	m->m_ifile = ifile;
 	m->m_scrpos.pos = pos;
 	m->m_scrpos.ln = ln;
+	free(m->m_filename);
 	m->m_filename = NULL;
 }
 


main.c
```diff
@@ -140,7 +140,11 @@ main(argc, argv)
 
 	s = lgetenv(less_is_more ? "MORE" : "LESS");
 	if (s != NULL)
-		scan_option(save(s));
```
gwsw, line 143: I'm wondering why the save() is here at all. Why don't we just call ```scan_option(s)```?

mark.c
```diff
@@ -59,6 +59,7 @@ cmark(m, ifile, pos, ln)
 	m->m_ifile = ifile;
 	m->m_scrpos.pos = pos;
 	m->m_scrpos.ln = ln;
+	free(m->m_filename);
```
gwsw, line 62: I don't believe m_filename can ever be non-NULL at this point. m_filename is set only in restore_mark, and cleared in mark_check_ifile when new_ifile is called. I don't see any way that m_filename can be non-NULL when cmark is called. Did you find such a case?

stoeckmann: I think it is only possible with a modified .lesshst file:
```
.less-history-file:
.mark
m a 1 0 /dev/null
m a 1 0 /dev/null
```
The cmark call would be done twice for the same character. I am not aware of a situation in which this can be achieved with a regular use case of less, but since free(NULL) is a no-op and it fixes a situation with such a .lesshst file I thought it won't hurt to add the free call.

gwsw:
Thanks. Fixed in 1974e5f8cd628e5fc1075883f3eba3e5390860d3 and 31a14b8124e551f9028c5a2785fbbcb5839c491d. I prefer to check for NULL before calling free() because there was some system years ago that crashed on free(NULL). It's probably no longer an issue but better safe than sorry.

PR #211, opened 2021-10-03
Fix floating point exception in protochar

stoeckmann:
If --header is set to "LINES - 1", i.e. one line less than
the screen height, then a "%d" in a prompt is not resolved
properly but crashes less instead.

How to reproduce (assuming your screen height is 25):
yes y | less --header 24 -P '%d'


gwsw:
Fixed in 20d1a3d50ab8f87e45476be9b33b151ffa4d36b3

PR #227, opened 2021-12-04
Fix signal race in iread.

stoeckmann:
The variable reading is used to indicate that ^C and other signals shall
"interrupt" a read call by performing a long jump. The jump target is
set within iread function.

Make sure that reading is always set back to 0 when iread is left, since
jumping to the target after leaving the target function is undefined
behavior.

This can be triggered on Linux systems (very hard to reach without
a debugger or another way to have enough time at the right spot):

term1$ mkfifo fifo
term2$ (yes y | head -n 50; read) > fifo
term1$ gdb less
\> break null_line
\> run -f fifo
[press F in less to set ignore_eoi to 1]
[kill fifo feeding process in term2, e.g. CTRL+C]
\> signal SIGINT

### DIFF
--- a/os.c
+++ b/os.c
@@ -158,7 +158,10 @@ start:
 		FD_ZERO(&readfds);
 		FD_SET(fd, &readfds);
 		if (select(fd+1, &readfds, 0, 0, 0) == -1)
+		{
+			reading = 0;
 			return (-1);
+		}
 	}
 #endif
 #if USE_POLL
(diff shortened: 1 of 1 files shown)


gwsw:
Good catch. But wouldn't it be simpler to move the ```reading = 1``` to just before the call to ```read()```, and move the ```reading = 0``` to just after the ```read()```? I don't think ```reading``` needs to be set except during the ```read()``` call itself.

stoeckmann:
If the only blocking call is read, then yes. I was not sure about the non-Linux code path because my only testing system was Linux.

gwsw:
I believe the ```select()``` in the DJGPP build can indeed block, so it's safer to do it your way. I made a small simplification and commited this in be39bf8585af74b8b8a1fd0e1a0937dd993e7259.

PR #229, opened 2021-12-06
Document lesskey command actions

codesoap:
I don't consider this pull request mature, but I thought it would be best to show you what I'm talking about, rather than trying to describe it.

When reading the lesskey man page, I felt like there was a missing section. This was written in it:

> The action is the name of the less action, from the list below.

But there is no such list in the man page. There is only an example, but no explanation of what each action does. Of course, one could try to open the less and lesskey man pages side-by-side and figure out like that, what each action does, but I feel like this is quite tedious. I would prefer there to be a list of all actions, with a description of what they do.

The `debug` action is undocumented right now and could also be documented in the lesskey man page (I have just added the placeholder "TODO" in this pull request).

There are also aliases for some actions, that are undocumented right now, e.g. `flush-repaint`, `end`, `toggle-flag`, ...

----

For this pull request I have simply copied the documentation from the less man page to the lesskey man page and adapted it a little bit to reference other actions rather than less commands. Maybe, however, something else would be better. Maybe a more brief description and a hint to look at the less man page for more info would be better.

### DIFF
--- a/lesskey.nro.VER
+++ b/lesskey.nro.VER
@@ -1,7 +1,7 @@
 '\" t
 .TH LESSKEY 1 "Version @@VERSION@@: @@DATE@@"
 .SH NAME
-lesskey \- specify key bindings for less
+lesskey \- customize key bindings for less
 .SH "SYNOPSIS (deprecated)"
 .B "lesskey [\-o output] [\-\-] [input]"
 .br
@@ -42,9 +42,9 @@ The input file consists of one or more
 Each section starts with a line that identifies the type of section.
 Possible sections are:
 .IP #command
-Defines new command keys.
+Customizes command key bindings.
 .IP #line-edit
-Defines new line-editing keys.
+Customizes line-editing key bindings.
 .IP #env
 Defines environment variables.
 .PP


gwsw:
I understand this concern, but I fear that proceeding along this path will result in this section of the ```lesskey``` man page being a copy of the COMMANDS section of the ```less``` man page. My original idea was that if someone was unsure about what a particular lesskey action does, they could look up the command key in the ```less``` man page for a full description. I understand that this is not ideal, but duplicating the same documentation in two different man pages also doesn't seem ideal or very maintainable. Perhaps adding a note to the lesskey man page about how to find the command descriptions in the less man page would be better?

codesoap:
After thinking a bit more about it, I agree that copying the whole section is probably not the best solution. Maybe I should take a step back and explain what my motivation was:

I'm thinking about writing a simple [Gemini](https://gemini.circumlunar.space/) browser and wanted to find out if `less` could be (ab)used as a "makeshift framework" for that. Thus I started reading the `less` man page, which referenced the `lesskey` man page. I initially thought, that `lesskey` could be used to somehow extend the capabilities of `less` by executing shell commands when a key is pressed or something similar.

Now that I have familiarized myself a little more with `lesskey`, I understand that this initial idea is not what `lesskey` is about. Instead it is just to be used for changing the default key bindings. Maybe if this was made clear at the beginning of the `lesskey` man page, I would have understood earlier, that I won't find any "new" actions (actions that are not bound by default) in there. Now that I know this, my desire for an action-list is not as great as before. I'll think about it a little more and update this pull request, if I can come up with a different suggestion...

polluks:
@codesoap Gemini & less? Already done https://tildegit.org/acdw/bollux

codesoap:
@gwsw: Sorry for the late response. I thought about this issue for a while but only came up with a small suggestion. The change clarifies a little bit, that `lesskey` is used to customize the default key bindings, as opposed to being able to add completely new functionality to `less`. I also, roughly, suggested this to the OpenBSD fork of less and it was accepted there: https://cvsweb.openbsd.org/cgi-bin/cvsweb/src/usr.bin/less/lesskey.1.diff?r1=1.17&r2=1.18&f=h

@polluks Cool, thanks, I hadn't seen this browser. I'll check it out!

PR #233, opened 2021-12-20
Fix stack buffer overflow

stoeckmann:
The adjustment of parse_error introduced a possible stack buffer
overflow if lesskey filename is long enough.

How to reproduce:

KEYFILE=$(python -c 'print("/tmp/"+4*(255*"a"+"/")+"lesskey.txt")')
install -d $(dirname KEYFILE)
echo "#version a" > $KEYFILE
less --lesskey-src=$KEYFILE -f /dev/null


stoeckmann:
Two notes:
- Latest stable release does not contain this bug
- This code uses snprintf directly, but SNPRINTF macros exist in less.h. Maybe it is time to switch code base to more robust snprintf usage

PR #234, opened 2021-12-20
Fix overflows with malformed lesskey lines

stoeckmann:
If lines are malformed it is possible to trigger out of boundary
read accesses during parsing.

Carefully handle the pointer increments to not move behind the
terminating nul byte.

How to reproduce:

python -c 'print(1022*" "+"\\")' > lesskey-1.txt
python -c 'print(1022*" "+"^")' > lesskey-2.txt
python -c 'print(1021*" "+"\\k")' > lesskey-3.txt

Open these files with lesskey, compiled with -fsanitize=address.

### DIFF
--- a/lesskey_parse.c
+++ b/lesskey_parse.c
@@ -181,6 +181,18 @@ char_string(buf, ch, lit)
 }
 
 /*
+ * Increment char pointer by one up to terminating nul byte.
+ */
+	static char *
+increment_pointer(p)
+	char *p;
+{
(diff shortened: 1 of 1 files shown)


stoeckmann:
I did not fuzz possible lesskey files so I do not know if this covers all cases. The last adjustment is probably not necessary -- but I added it just in case.

For style reasons it might make sense to modify all *pp = p+1 lines in that function.

Issue #248, opened 2022-01-17
OSC-8 search


sdaoden:
Hello!
In 2014 i sent an enhancement via mail that added support for a "mdocmx" groff extension with which less could become an interactive browser.

After less gained OSC-8 support in v566 i rewrote all that and based it upon OSC-8 instead of extending (aka misusing) the traditional backspace Unix manual formatting approach for this.  Thanks for this support!
I also opened a groff extension request, and it was actually implemented (https://savannah.gnu.org/bugs/?60666), and will thus be in the next release of groff, whenever it will happen.
I sent an according patch last year, but got no response.

Please find at [1] a patch against less v600 (11610 bytes) at your disposal.
It adds searching for OSC-8 sequences;  it performs a local search with scrolling for id='s, and with HAVE_SYSTEM it adds the possibility to start man(1) for special man:// URLs.
It assumes OSC-8 sequences (the sequence itself) stay on a line, and i think i recall that less(1) does not handle them across line boundaries either.

  [1]  https://ftp.sdaoden.eu/less-osc8-search.patch

Thanks for consideration.

sdaoden:
P.S.: the manual of the MUA i maintain uses mdocmx sequences, this still requires groff extensions which are not present to extend the mdoc macros to generate anchors and links.  But you can interactively browse a manual on a Unix console with it.  For the interested i pre-create them at release time of this MUA, at [1] you could find the latest, for example.

 [1] https://ftp.sdaoden.eu/s-nail-14.9.23.xcat1.xz

sdaoden:
P.P.S: like in, #id == search local, otherwise pass to LESSURLOPEN if set in environment and secure mode is not on.
(And sorry i am new to github .. and came only (back) to report issues to lessen people's burden with me, you know, they block some countries for reasons that i find odd, especially in context. I am reading about regular pull requests..)

sdaoden:
Hello!
So i looked into this, i was maybe a bit in overreaction since the two mails in 2014 and last year i think did not get any response ..., but i think this can be done much better, without the focus on mdocmx i wrote in 2014.

[...]

PR #251, opened 2022-01-29
Add ^O series to handle OSC-8 id= search and URIs

sdaoden:
As promised in Issue #248, here is a pull request to really and truly handle OSC 8 links with all its powers, instead of just that 2014 hack to make the mdocmx roff macro references work on the terminal, adjusted to use OSC 8.

Even though just finished an hour ago .. i think it is working fine, with the exception as below.

 . ^O^I text  -  will search for the OSC 8 id= "text".
 . ^O^N  -  searches for the next OSC 8 link aka URI.
 . ^O^P  -  searches for the previous OSC 8 link aka URI.
 . ^O^O  -  opens the currently selected OSC 8 link aka URI with the shell command given in the environment variable LESSOSC8OPEN; it will be passed as a properly quoted single argument.
                If LESSOSC8OPEN is not set, "man:NAME(\(.*\))?" style links are still understood and opened via man(1).

The good intent of this is reflected by the fact that search.c now has 2442 lines!

So i had prepared a test file with bogus entries and such, and it works very nicely.
I for one see only two problems:

 . in the status line when you see those ~ lines because we scrolled "past EOF" there will be input artifacts, ie the repainting is somehow not clean, _if_ you ^O.. ^O.. like grazy at least.
   That is to say that those ~ lines may look like "~^O^N", i do not know why.
   It is a repainting issue of course.

 . i cannot figure out how to configure the search so that for ^O^N and ^O^P do not wrap but get the first / the last link right; for example, in a test file with 34 id=s, i ^O^I34, and (it has a link to, not only an id aka anchor) ^O^N will say "No OSC 8 link found  (press RETURN)", but if i then ^O^N again, it will find the links in the last line again, it does not step over it.
  Likewise for the first anchor, if i ^O^P, then the search wraps backwards and starts at EOF again.

Other than that .. it is fun!  Browsing on the terminal, but in a generic way, not like i did in 2014 with mdocmx only for mdoc man(1)uals using crude hacks!!

I hope you like it!
..and a nice Sunday i wish from Germany.

### DIFF
--- a/cmd.h
+++ b/cmd.h
@@ -77,6 +77,11 @@
 #define A_X116MOUSE_IN         68
 #define A_PSHELL               69
 #define A_CLR_SEARCH           70
+#define A_OSC8_ID_SEARCH       84
+#define A_OSC8_ID_SEARCH_LOAD  85
+#define A_OSC8_GOTO_P          86
+#define A_OSC8_GOTO_N          87
+#define A_OSC8_OPEN            88
 
 /* These values must not conflict with any A_* or EC_* value. */
 #define A_INVALID              100
--- a/command.c
+++ b/command.c
@@ -333,6 +333,12 @@ static void exec_mca(void)
 			error(done_msg, NULL_PARG);
 		break; }
 #endif
+#if OSC8_SEARCH
+	case A_OSC8_ID_SEARCH:
+	case A_OSC8_ID_SEARCH_LOAD:
+		osc8_id_search(cbuf, mca == A_OSC8_ID_SEARCH);
+		break;
+#endif
 	}
 }
 
@@ -2155,6 +2161,32 @@ public void commands(void)
 
 		case A_NOACTION:
 			break;
+
+#if !OSC8_SEARCH
+		case A_OSC8_ID_SEARCH:
+		case A_OSC8_ID_SEARCH_LOAD:
+		case A_OSC8_GOTO_P:
+		case A_OSC8_GOTO_N:
+		case A_OSC8_OPEN:
+			error("Command not available", NULL_PARG);
+			break;
+#else
+		case A_OSC8_ID_SEARCH:
+		case A_OSC8_ID_SEARCH_LOAD:
+			start_mca(action, "[OSC 8 id]:",
+				(void *)NULL, CF_QUIT_ON_ERASE);
+			c = getcc();
+			goto again;
+		case A_OSC8_GOTO_P:
+			osc8_goto(SRCH_BACK);
+			break;
+		case A_OSC8_GOTO_N:
+			osc8_goto(SRCH_FORW);
+			break;
+		case A_OSC8_OPEN:
+			osc8_open();
+			break;
+#endif /* OSC8_SEARCH */
 
 		default:
 			bell();
--- a/configure.ac
+++ b/configure.ac
@@ -561,6 +561,11 @@ AH_TOP([
 #define	LOGFILE		(!SECURE)
 
 /*
+ * OSC8_SEARCH is 1 if you wish to support searching/following OSC8 references.
+ */
+#define	OSC8_SEARCH	(!SECURE)
+
+/*
  * GNU_OPTIONS is 1 if you wish to support the GNU-style command
  * line options --help and --version.
  */
--- a/decode.c
+++ b/decode.c
@@ -179,7 +179,12 @@ static unsigned char cmdtable[] =
 	'Q',0,                          A_QUIT,
 	':','q',0,                      A_QUIT,
 	':','Q',0,                      A_QUIT,
-	'Z','Z',0,                      A_QUIT
+	'Z','Z',0,                      A_QUIT,
+	CONTROL('O'),CONTROL('I'),0,    A_OSC8_ID_SEARCH,
+	CONTROL('O'),CONTROL('L'),0,    A_OSC8_ID_SEARCH_LOAD,
+	CONTROL('O'),CONTROL('N'),0,    A_OSC8_GOTO_N,
+	CONTROL('O'),CONTROL('P'),0,    A_OSC8_GOTO_P,
+	CONTROL('O'),CONTROL('O'),0,    A_OSC8_OPEN
 };
 
 static unsigned char edittable[] =
--- a/defines.ds
+++ b/defines.ds
@@ -98,6 +98,11 @@
 #define	LOGFILE		(!SECURE)
 
 /*
+ * OSC8_SEARCH is 1 if you wish to support searching/following OSC8 references.
+ */
+#define	OSC8_SEARCH	(!SECURE)
+
+/*
  * GNU_OPTIONS is 1 if you wish to support the GNU-style command
  * line options --help and --version.
  */
--- a/defines.o2
+++ b/defines.o2
@@ -91,6 +91,11 @@
 #define	LOGFILE		(!SECURE)
 
 /*
+ * OSC8_SEARCH is 1 if you wish to support searching/following OSC8 references.
+ */
+#define	OSC8_SEARCH	(!SECURE)
+
+/*
  * GNU_OPTIONS is 1 if you wish to support the GNU-style command
  * line options --help and --version.
  */
--- a/defines.o9
+++ b/defines.o9
@@ -90,6 +90,11 @@
 #define	LOGFILE		(!SECURE)
 
 /*
+ * OSC8_SEARCH is 1 if you wish to support searching/following OSC8 references.
+ */
+#define	OSC8_SEARCH	(!SECURE)
+
+/*
  * GNU_OPTIONS is 1 if you wish to support the GNU-style command
  * line options --help and --version.
  */
--- a/defines.wn
+++ b/defines.wn
@@ -91,6 +91,11 @@
 #define	LOGFILE		(!SECURE)
 
 /*
+ * OSC8_SEARCH is 1 if you wish to support searching/following OSC8 references.
+ */
+#define	OSC8_SEARCH	(!SECURE)
+
+/*
  * GNU_OPTIONS is 1 if you wish to support the GNU-style command
  * line options --help and --version.
  */
--- a/less.h
+++ b/less.h
@@ -347,6 +347,14 @@ struct wchar_range_table
 	unsigned int count;
 };
 
+/* State while processing an ANSI escape sequence */
+struct ansi_state
+{
+        int hindex;   /* Index into hyperlink prefix */
+        int hlink;    /* Processing hyperlink address? */
+        int prev_esc; /* Prev char was ESC (to detect ESC-\ seq) */
+};
+
 #if HAVE_POLL
 typedef short POLL_EVENTS;
 #endif
@@ -395,6 +403,9 @@ typedef short POLL_EVENTS;
 #define SRCH_SUBSEARCH(i) (1 << (16+(i))) /* Search for subpattern */
 /* {{ Depends on NUM_SEARCH_COLORS==5 }} */
 #define SRCH_SUBSEARCH_ALL (SRCH_SUBSEARCH(1)|SRCH_SUBSEARCH(2)|SRCH_SUBSEARCH(3)|SRCH_SUBSEARCH(4)|SRCH_SUBSEARCH(5))
+#if OSC8_SEARCH
+# define SRCH_OSC8	(1 << 22) /* OSC 8 id= search */
+#endif
 
 #define SRCH_REVERSE(t) (((t) & SRCH_FORW) ? \
                                 (((t) & ~SRCH_FORW) | SRCH_BACK) : \
@@ -625,7 +636,6 @@ typedef enum {
 struct mlist;
 struct loption;
 struct hilite_tree;
-struct ansi_state;
 #include "pattern.h"
 #include "xbuf.h"
 #include "funcs.h"
--- a/less.hlp
+++ b/less.hlp
@@ -75,6 +75,11 @@
 
   m_<_l_e_t_t_e_r_>            Mark the current top line with <letter>.
   M_<_l_e_t_t_e_r_>            Mark the current bottom line with <letter>.
+  ^O^I_<_i_d_>             Search for an OSC 8 <id> in the file.
+  ^O^P                 Go to previous OSC 8 link.
+  ^O^N                 Go to next OSC 8 link.
+  ^O^O                 Open current OSC 8 link with LESSOSC8OPEN.
+  ^O^L_<_i_d_>             Combined ^O^I plus ^O^O.
   '_<_l_e_t_t_e_r_>            Go to a previously marked position.
   ''                   Go to the previous position.
   ^X^X                 Same as '.
--- a/less.nro.VER
+++ b/less.nro.VER
@@ -105,6 +105,17 @@ and LEFTARROW commands.
 Scroll horizontally right to show the end of the longest displayed line.
 .IP "ESC-{ or \(haLEFTARROW"
 Scroll horizontally left back to the first column.
+.IP "^O^I"
+Search for an OSC 8 ID in the file.
+.IP "^O^P"
+Go to the previous OSC 8 link.
+.IP "^O^N"
+Go to the next OSC 8 link.
+.IP "^O^O"
+Open the current OSC 8 link with the command specified
+in the LESSOSC8OPEN environment variable.
+.IP "^O^L"
+Combined ^O^I plus ^O^O.
 .IP "r or \(haR or \(haL"
 Repaint the screen.
 .IP R
@@ -2316,6 +2327,11 @@ Enables individual features which are normally disabled by LESSSECURE.
 See discussion under SECURITY.
 .IP LESSSEPARATOR
 String to be appended to a directory name in filename completion.
+.IP LESSOSC8OPEN
+Command line to invoke when following OSC 8 links.
+Without this being set only
+.IR man (1)
+references like "man:ls(1)" can be opened.
 .IP LESSUTFBINFMT
 Format for displaying non-printable Unicode code points.
 .IP LESSUTFCHARDEF 
--- a/lesskey.nro.VER
+++ b/lesskey.nro.VER
@@ -179,6 +179,11 @@ g	goto-line
 \ekh	goto-line
 <	goto-line
 \ee<	goto-line
+^O^I	osc8-id-search
+^O^L	osc8-id-search-load
+^O^N	osc8-goto-next
+^O^P	osc8-goto-prev
+^O^O	osc8-open
 p	percent
 %	percent
 \ee[	left-scroll
--- a/lesskey_parse.c
+++ b/lesskey_parse.c
@@ -72,6 +72,13 @@ static constant struct lesskey_cmdname cmdnames[] =
 	{ "next-tag",             A_NEXT_TAG },
 	{ "noaction",             A_NOACTION },
 	{ "no-scroll",            A_LLSHIFT },
+#if OSC8_SEARCH
+	{ "osc8-id-search",       A_OSC8_ID_SEARCH },
+	{ "osc8-id-search-load",  A_OSC8_ID_SEARCH_LOAD },
+	{ "osc8-goto-next",       A_OSC8_GOTO_N },
+	{ "osc8-goto-prev",       A_OSC8_GOTO_P },
+	{ "osc8-open",            A_OSC8_OPEN },
+#endif
 	{ "percent",              A_PERCENT },
 	{ "pipe",                 A_PIPE },
 	{ "prev-file",            A_PREV_FILE },
--- a/line.c
+++ b/line.c
@@ -135,13 +135,6 @@ static struct color_map color_map[] = {
 	{ AT_COLOR_SUBSEARCH(5),   "Wc" },
 };
 
-/* State while processing an ANSI escape sequence */
-struct ansi_state {
-	int hindex;   /* Index into hyperlink prefix */
-	int hlink;    /* Processing hyperlink address? */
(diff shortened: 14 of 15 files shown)


sdaoden:
P.S.: that web form seems to not like reverse solidus aka backslash, i do not know how to write the regular expression correctly.  Sorry.

sdaoden:
Hello.  I am looking into this since yesterday, and repainting was really, really bad.  It is very much better now locally, but i still have some problems with ^O^N and ^O^P repositioning.  I hope i get this fixed until tomorrow, and i think then this pull request had been polished even without voice of "the expert".

I will push it then, and "try to" attach a nice grotty(1) manual file for testing purposes!

Oh, i hope you like it then.
Thank you.

sdaoden:
With "Add shared static osc8_search_xy(), properly restore search pattern.." i think we now really handle movement and repaint graceful (expensive though).

I still do not use the builtin shell_quote because of our specific needs.  This could of course be changed, it would likely look better in almost all cases (when user is prompted to comply visiting external manual etc).

One problem is left, a sole "#" URI without any OSC 8 parameter.  When then using ^O^O to follow it, it will not "find the pattern".  This could be special cased, but seems unlikely in practice.

Unfortunately i cannot attach the file!  Not as tgz gz whatever, they say "We don't support that file type".
So i stored the contained converted manual, to be read via "less -RIF", on my vserver as

$ curl -I https://ftp.sdaoden.eu/z.cat.gz
HTTP/2 200
content-type: application/gzip
etag: "2778712001"
last-modified: Fri, 09 Sep 2022 17:18:51 GMT
content-length: 6353
accept-ranges: bytes
date: Fri, 09 Sep 2022 17:19:11 GMT
server: lighttpd/1.4.66

I think this is it now.  I hope you consider its inclusion in less.
Thank you.

sdaoden:
Works flawless for me.
It would be very nice if you would at least peep, i am trying to contact you since 2014 on this, or rather the preceding idea in the end.

gwsw:
This seems like a pretty big change for a rather niche feature. Couldn't you just temporary turn off -R if you want to search for text in an OSC8 sequence?

sdaoden:
Hm.  I would not call it a "niche feature".
Your idea is interesting.  The OpenBSD people added a tag file feature
for their mandoc replacement of man(1), which is more in the direction
i was thinking of.

The next release of groff _will_ bring support for OSC-8 generation in
grotty(1), and .HR is the new reference command called for man(7) macros
ithink.  So if converters / manual writers start using it, looking at
manuals in less(1) becomes a fully interactive experience.  At least you
can/could jump in between anchors and follow references with the new ^O
series of commands.

Maybe they get the taste of it then, finally.  And come back to the idea
from 2014 that is realized with the mdocmx(7) extender of the mdoc(7)
manual macros used for BSD-style manual pages -- because if you look at
what normal UNIX manuals _could_ offer in conjunction with less(1), they
would possibly not offer HTML-ified manuals no more (i think NetBSD for
example does), because it is simply not needed, you get the full
interactive experience as with GNU info (or even better) simply via
man(1).  Please let me point to three readily prepared "cat" pages that
can be simply used within a patched less(1), with no other programs
involved (below).

It is .. a big change, but the overall change in between the master and
the topic branch via wc(1) -Lwcl is not so large.  Ok it is large.
But it is not so large given that you pimp it up to a fully interactive
user experience.  Oh .. my god .. Letterman will invite me!!!  Eh, you!

        37964 132839 901211    180 total
        38554 134699 913950    180 total

Following external references via LESSOSC8OPEN or the builtin manual
thing could be stripped, but .. if you get used to it, it is nice: even
better than GNU info i think, you can open external manual pages and
then come back to where you started.  UNIX-like, just quit the new
manual process, and come back.  (Ie job-control-alike.  A bit.)

(Readily prepared manuals, simply for "$ less --RAW-CONTROL-CHARS
--ignore-case --no-init FILE"

        https://ftp.sdaoden.eu/code-mailx-1.cat

A huge one (~ 1MB) of my mailer, with hundreds of references.
And, much smaller, but ditto:

        https://ftp.sdaoden.eu/code-mdocmx-7.cat

Of the mdocmx extension of the roff mdoc manual macros.

        https://ftp.sdaoden.eu/code-mdocmx-1.cat

Of the preprocessor for it.)

P.S.: i have had to add another commit to the topic branch,
"osc8_search_xy(): call search() right or SEGV will happen" (changes the
"match" parameter from 0 to 1).

Thank you for your message!
I was trying to get in contact with you since 2014.
:-)

Ciao and a good and healthy 2023 i wish from Germany!

sdaoden:
Sorry to be so noisy.  I had seen you dropped K&R for ISO prototypes.  (I added four more commits, in order).

Thank you.  Ciao and greetings.

gwsw:
Can someone clarify something for me? I thought the idea of OSC-8 links was that the terminal handles them. You click on them or something to navigate to the link. Why does less need to do anything other than pass the sequence to the terminal?

sdaoden:
Hm. Pushing again as i have seen that i re-brought in the get_seg() function that was gone in less version 244 if that look was correct.  Also some "constant" stuff (step_char() is not yet constant in comparison to step_charc() which already is; but could; anyhow).
I folded all changesets into one.

OSC8 on the terminal, yes.
But with this patch you can use O^N and O^P to scroll to these anchors in the text, and with O^I you can search for IDs, and with O^O you can open the "current" anchor (via $LESSOSC8OPEN if it is not handled internally -- then after user confirmation, and not in safe mode, iirc).)

If people realize what they get, it could be they start generating manual pages (or converters which do that, effectively) etc which use this feature.  You get a notion as via browsing a HTML document in a text mode browser like lynx.

I still have the manuals available via download.  They were generated via my own mdoc macros for UNIX manuals, which the groff maintainer did not accept for upstreaming, but still they show what is possible.  Because they base upon upstream mdoc macros, we only can apply [NUMBER] tags to announce anchors/URLs, not prepend them like lynx(1) does, nor simply enwrap the actual "URL text" as is done by graphical browsers.  (Ie, "paint it blue", or the like.)

Ie, look at with patched less -RIFe, scroll point-of-interest into view with ^O^I, scroll back with '', just as usual.

Small file: 
  https://ftp.sdaoden.eu/code-mdocmx-1.cat
Medium:
  https://ftp.sdaoden.eu/code-mdocmx-7.cat
Large, with many hundreds of links:

sdaoden:
https://ftp.sdaoden.eu/code-mailx-1.cat

sdaoden:
P.S.: "current" GNU roff / groff generates the needed IDs, finally, and at least.

gwsw:
@sdaoden, I apologize for my lack of responsiveness to this issue. I have been studying your changes and I have one question. Can you explain the purpose of searching for an id value? From my reading, the main purpose of the id parameter is to allow terminal emulators to highlight the appropriate pieces of text on mouseover. Given that purpose, I don't see why a user would have a need to search for an id value.

sdaoden:
Sure, Mr. Nudelman.

To reiterate that the "standard" for OSC8 allows key=value pairs,
but only defines id=.  That year-long unmaintained document
describes that a text editor on-the-fly generates some kind of id
that can happen multiple times, but should always be identical for
the same data, so that the terminal emulator can, for example,
highlight all equal-IDs on a screen.

I personally always found that meaning of ID strange, in
comparison to UUID, to HTML #id or anchors, etc.  To TeX idx
(-alike macros).  Etc.  I have not seen that implemented myself,
either.

But *if* one extends the meaning of that ID to something
in-document aka permanent aka non-dynamically generated, to be
a real anchor, really like HTML's "a name=ID", then one can use
OSC8 to provide the fully interactive experience with in-document
anchors and references, too.

This can of course be anything, ie, strings, like a UNIX manual
section headline OPTIONS etc.  For my mdoc macros i followed the
lynx text browser, which enumerates all links in a HTML document.
Typing numbers is usually also shorter and less error prone (and
case-insensitivity is locale-dependend).

So in short.  In addition to the ^O^P and ^O^N, and ^O^O commands,
to hop around the OSC-8 references in a document, and open man:
directly and the reset via $LESSOSC8OPEN when set, giving the
opportunity to search for embedded IDs via ^O^I *can* give a full
interactive experience if the document makes use of it.

I linked example documents which give this fully interactive
experience, with table of content, and to-the-point scrolling.
For now i do not know any other roff macro package which makes use
of this, we will see how the Linux manual maintainer makes use of
the new grotty feature to create such references (and KEY=VALUE).

But i know people go to the web to look at HTMLified manual pages
with fewest links, to have the actual opportunity, (NetBSD, at
least in the past, also installed that locally on request if
i recall correctly), even though the manual pages are installed
locally and can be made as interactive, or even better, and
offline.  Again i can only point to the example documents.

Thus i for one would add that ^O^I, speculating it will happen for
real.  And make use of it in the less manual page, which is pretty
lengthy.  And preformatted!  So that would be very easy to do.
The necessary code snippet for the explicit search is quite short.
One needs to activate the ID search via shortcut command,
otherwise it is not in the way.

P.S.: it also has to be said that instrumenting manual page source
code with index anchors allows for really nice interlinked "manual
page books".  This is especially true for the mdoc language that
is natively used on BSDs, because it has semantic and hows about
macros, functions, constants, etc etc etc.
And these real indexes of mdoc, they can now be referenced to the
point.
Most projects i know have switched their manual page sources to
some other format in order to be able to get semantics, and other,
interactive / referencing document formats.
So these only need support of their converter, and ^O^I can spring
into existence with full power.

gwsw:
@sdaoden Thanks very much for the detailed explanation. When you refer to the "year-long unmaintained document" I assume you're talking about https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda. Is there any kind of formal RFC that specifies the OSC8 behavior other than that page? I'm a little leery about implementing features simply based on you or I "speculating" on how OSC8 might be used in the future. This might lock less into doing something that can't easily be changed due to backwards compatibility, but is in conflict with what actually happens in the future.

sdaoden:
I do not think there is anything else on the internet.  I think he is the originator of the OSC8 feature you have implemented in less(1).

To reiterate that ^O^[NPO] is unaffected of your doubts.

Regarding ^O^I alone i find your statement a bit strong.

Hm, i mean .. one could think of extending this even, you know, in making it regular, maybe naming it ^O/ instead (^O^S will not do), so that *any* user input is searched for only in OSC8 content, VALUEs of any given KEY=, as well as the link itself?
You know, a regular search in OSC8 constructs that covers any [KEY=]VALUE and link content, so that if i do for example "^O^I"less it will find the "next" OSC8 anchor/link which refers to "less"?
That strikes me as a good thing totally apart of any possible doubts on the adaption of "id="UUID?
How about this?

sdaoden:
.. this would be an "any substring matches" search then, affected by the usual less semantics of case-sensitive.  I had to look on how to implement this, then.

sdaoden:
(P.S.: to clarify: the one who "picked it up" is on the GNU groff mailing-list for some years.)

sdaoden:
Hello!  That is wonderful news!!  (Of course i am .. but only a bit.)
I will try to look into this today, only saw the commitdiff until now; quite a bit different :-)
Thank you!

sdaoden:
^O^L does not seem to do what is documented (anything, in fact).
But ^O^O does follow in-document #anchor links, so there i went.

I find the LESS_OSC8_xxx thing a bit excessive compared to my single "muxer" approach; in general i have the impression that "muxing" is preferred "by the people", but then again your omnipresent program has a multi-decade long history with its LESSOPEN mechanism, so if you prefer a different one now, then surely for a reason.

The manual says LESS_OSC8_xxx but has LESS_OSC_man and LESS_OSC_file typos.

I am very happy that you provide a built-in LESS_OSC8_man -- this is very, very cool, and i really look forward that manual pages make use of this feature!  It is, despite for the internal #anchor links, what i was hoping for since 2014!  You get a completely interactive UNIX manual session simply by following a link, on a normal terminal!  Yes.

Ah wait.  If i follow an OSC-8 #anchor via ^O^O, and then ^O^L, we scroll back to the OSC-8 which was followed.  Hm.  Well less(1) has '' to jump back where you were last, but you surely had a use case idea.

For me on the dark side is that with that lynx-alike manual page extension i have written, which provides that "numbered link view" that lynx invented, as far as i know, i now have to count links on the current screen, to N^O^N there to ^O^O thereafter, but, well, it is a specific thing.
Then again, without such visualization, the "N" of N^O^N is hardly known.
But i agree you are right when you say that support for that is not insight, and rather people will, like mandoc(1) does when converting UNIX manuals to HTML, create anchors like "#INPUT_PREPROCESSOR" (or even "#INPUT PREPROCESSOR") than anything else.

^O^N and ^O^P do not "wrap around" from the bottom to the top and vice versa.  Which i got used to.  Likely personal taste only.

Dear Mr. Nudelman, thanks for bringing this in.
Now i hope for people creating an almost (plain) webpage-alike UNIX manual experience.
Or even normal reading in textualized PDFs.

gwsw:
@sdaoden thanks for your comments.

Regarding LESS_OSC8_xxx, my thinking is someone might want to change, for example, their browser (which opens https: links), or their file viewer (which opens file: links) but in general when they do something like that, they want to leave handling of other types of links unchanged. It seems easier to manage this if the handler for each type of link is separate, rather than having to edit a muxer script that handles many different types of links. You could of course set several LESS_OSC8_xxx variables to point to the same script. But perhaps it would be convenient to have a LESS_OSC8_all or something like that, which would handle all link types that don't have a specific handler.

^O^L was a late addition, and I'm not positive it's really needed, but sometimes I would follow an internal link, then navigate around for a while, then want to return to the original link. The navigation may have overwritten the place where '' (quote-quote) would return, so ^O^L seemed somewhat useful.

I agree counting links in order to supply a number to ^O^N is not ideal. I'm thinking of allowing link navigation by just clicking on the link with the mouse, when --mouse is enabled.

sdaoden:
Hello.  ^O^L i simply have misunderstood it seems.  I think if a fresh mind reads this she or he will get that right immediately.  I .. was using local modifications for almost ten years and got used to my flow.

I can absolutely follow and agree with your thoughts on the individual handlers. (I have never used lessopen myself; i think somewhen "decades ago" FreeBSD activated lesspipe generically, but this is only dim memory -- installed it still is.)

Mouse on the terminal i have not used for long (i disable anything but select and paste in the st(1) i use), so i cannot comment on that.
But if you think of such, maybe a shortcut to highlight all OSC8 links (not anchors) in (the visible part of) a document would be of interest?

Btw i now (sorry for my short breath, i have to finish a DKIM thing until February 1st before Google requires it, and just today finished the RFC 5322 parser) see that ^O^O seems to perform work on an OSC-8 anchor, ie, an OSC8 link without URI target.
I think it would be better if those anchors-only things would simply be skipped; iirc my patch did so.
(I have switched to your regular implementation!)

sdaoden:
I said

  But if you think of such, maybe a shortcut to highlight all OSC8 links (not anchors) in (the visible part of) a document would be of interest?

If you want to spend real time that "highlight all OSC8 URIs on the screen" mode could actually glue a generated "N" to the URI and highlight only that.  That is a lot of work, but then, at least on the current screen, less(1) would act almost like lynx(1) does.
That is, then, i "highlight all OSC8 links on the screen", and then can do an exact "N"^O^N because i do not have to count, plus ^O^O to go where i want!
Of course doing this will reflow the entire screen.

gwsw:
> ^O^O seems to perform work on an OSC-8 anchor, ie, an OSC8 link without URI target.
I think it would be better if those anchors-only things would simply be skipped; iirc my patch did so.

Can you give a specific example of this? As far as I know, ^O^N ignores OSC8 sequences where the URI or the marked text is empty. So ^O^O shouldn't operate on them if they aren't selected by ^O^N. For example, in your code-mailx-1.cat file, ^O^N at the beginning of the file goes to the NAME link on line 24, and ignores the three anchors on lines 3, 6 and 23.

sdaoden:
Hello!  No, it does not, actually.
The files i offer are all non-debug, but the mdocmx macros have a debug mode so that you can see anchors.  They are still only anchors, but there is text in between the opening OSC8 sequence with URI (and ID), and the closing empty one.

The difference in between my patch and yours is that your ^O^N interprets these anchor-only OSC8 sequences as links.
I want to remark that i define "anchor-only" those which have no URI in the sequence, as opposed to no-link-text.  Unfortunately there *has* to be some link-target, i use "#" number sign, because the groff maintainer implemented the grotty command that way; empty URI, even only two parenthesis (last i tried) does not work out, which is very, very unfortunate.

So i used a plain number-sign "#" as an "anchor URI".

[debug.txt](https://github.com/gwsw/less/files/14069220/debug.txt)
[ndebug.txt](https://github.com/gwsw/less/files/14069222/ndebug.txt)

sdaoden:
Ah, this works!  Minimal examples these are, the ndebug variant has only one real link, and with my version the debug variant has, too, whereas with yours you will have three.

Yes, i did not think about my environment before posting my complain.
But i want to note that quite often IDs / anchors are applied to non-empty text.
This is not true for the plain mdoc(mx) aka roff input that produced the examples, as there placed the in-text anchors consciously "somewhere in the middle of a paragraph" to make the scrolling via man(1) catch up correctly (as of experience).  (I can show examples in the mailer's manual upon interest.)

But in HTML or other such formats an ID or anchor is often attached to a for example <span> or other such element, which then has textual content (or an , for example, but also, in the case i want that landing at the anchor gets not only the image but the surrounding paragraph).
In that scenario, if an automatic converter (shall any be updated to support such OSC8 links) creates plain text with an anchor, your approach could practically highlight an entire paragraph.
Well, at least possibly.
It is very unfortunate that the groff maintainer implemented the thing the way he did.

sdaoden:
P.S.: but you can see, that ^O^N finds the "anchor-only", and ^O^O then searches and says "OSC8 link not found".
And i think there needs to exist a possibility for an "anchor-only".
The OSC8 spec surely allows that.

sdaoden:
P.P.S.: eh!  Of course this browser-thing mis-interpreted the things, i had written HTML SPAN and IMG markup tags.

sdaoden:
P.P.P.S.: ;)
Please note the difference in between debug and non-debug is only the link text in between the two OSC8 sequences that make up a link, the URI is the same only-number-sign, and the ID= is, too.

sdaoden:
P.P.P.P.S.: on the other hand your approach is just another way of doing things.  And converters can, if they do it, simply generate an appropriate -- to your approach -- OSC8 link.
I think i simply change my thing to comply, the groff/grotty change is out as-is and cannot be helped, anyway.

What do you think about the "highlight all links" idea?
For users of a terminal, or directfb2, etc, there will be no super-duper terminal which does so.

gwsw:
In 846bc3005d4634b2acbd7635b4227bcb79cecfa9 less ignores URIs consisting of a single "#". Although I'm a little uneasy that we seem to be inventing a spec ourselves here.

I'm not sure what the user experience would be with the proposed "highlight all links" command. If you highlighted all links and then pressed ^O^O, what would happen? Would it open the first one on the page? I guess I'm not fully understanding this proposal or what it solves.

sdaoden:
Hello!  Yes, it was unfortunate how it was done, i tried what i could; i then unsubscribed from the groff lists.
Anchor-only aka "here is index" must be possible *somehow*.  And if "#ANCHOR" defines an anchor ANCHOR, then "#" does not define an anchor at all, i thought, and think that can be understood.
I had changed my mdocmx macros right away after my messages, so an empty link content is now always generated for anchors.
But i do support your change, obviously.

Regarding "if you highlighted all links and then pressed ^O^O, what would happen?"
I did not even think about this.  I would think it is unrelated with each other?
Then again, and if recall correctly, it would be implemented by a search and likely the first match is "hot", so to say, when the user comes into play again.  I would thus understand if ^O^O would then open the topmost OSC8 link on the current screen?

I was only thinking that on a normal console without specific OSC8 support, and unless the document is prepared like my example files that you know, with "links standing out in blue", for example, one cannot the links at all.  One can hop ^O^N etc link by link, or with trial and error do "10^O^N", for example.
But if there would be ^O^H or the like (backspace?), there would be an immediate impression on what there is on the visible screen, and getting to the link of interest would be faster i presume (especially with a small number of links on the visible screen a command like for example "4^O^N" should come quite easy, then).

(What would need screen reflow and lots of work i think would be that less(1) does something like lynx(1) does then: lynx places a "[NUMBER]" tag right before each href, say, "[3]", and if you then type "3" it says "Follow link (or goto link or page) number: 3", and you only need to hit RETURN to get there.  This would likely be the super-luxury variant of a ^O^H :-)

Then again i admit that i personally never used a terminal with real OSC8 support, but i presume most people do, so if the terminal already highlights OSC8 links on the screen the functionality as such, but especially the super-luxury variant, is possibly overkill.

But, you know, this is where i came from.  I came from these lynx(1) user experience impressions, and just wanted to mimic it for plain text.  And since 2014 i can look at my own manuals (and did some for NetBSD in the meantime, to get some traction for the idea, and they did install HTML manuals, but, well; this was pre-OSC8, though), and i read through the text, and then simply typed ^O^I and a number, and got there.

But i am super happy with that less has now; it is just that, you know, if i see the table of contents, and have to ^O^N repeatedly, or "do binary search", it, well, it sucks a bit.  For example,

     NAME[1]
     SYNOPSIS[2]
     TABLE OF CONTENTS[3]
     DESCRIPTION[4]
           Options[5]
           A starter[6]
           On sending mail, and non-interactive mode[7]
           Compose mode[8]
           On reading mail, and more on interactive mode[9]
           HTML mail and MIME attachments[10]

Without the numbers already such a small and sequential and one-per-line thing is hard for i think most human brains, regarding "NUMBER^O^N".
The idea of that "highlight all links" would however be to make them visually stand out (reverse video etc) on the screen, to make that easier.

Ciao!

(P.S.: i only threw away my topic branch, i was unaware that the issue is then "closed"..)

PR #267, opened 2022-06-07
Mention CR display behaviour before LF

SuibianP:


gwsw:
Man page clarified in e1c0c359997a319bb68842190859a85a473e759c.

PR #284, opened 2022-09-28
Smarter return codes

polluks:
More flexible for systems like AmigaOS.


gwsw:
Implemented slightly differently in c64e51b53fca8ab519a7851a16432ae0d7fe620c.

PR #296, opened 2022-10-25
Better Perl portability

polluks:
```
$ which perl
/usr/pkg/bin/perl


polluks:
I missed https://github.com/gwsw/less/blob/71bbae09c74f68ba47933dbc1e358a542a6f8f03/lesstest/runtest#L1

gwsw:
Fixed runtest and maketest in 5edb416bf72bb96d09fa7f93da93647fc69e1b91.

Issue #258, opened 2022-04-02
Input preprocessor failures are not diagnosed


eggert:
Currently if the input preprocessor fails, `less` does not report this failure to the user. This causes problems with applications like Gzip's `zless` program, which runs `less` with `LESSOPEN='||-gzip -cdfq -- %s'` in the environment, and if `gzip` fails no failure indication is given to the user. I found this out a while ago via Jaroslaw Weglinski's `gzip` bug report "[zless error under ulimit -v](https://bugs.gnu.org/16951)" but haven't gotten around to trying to fix it until now.

As I mentioned in that bug report, you can reproduce the problem by creating a large text file `t` and compressing it with `gzip` (say, by `seq 1000000 >t; gzip t`), and then by running the following shell command in one window:

`LESSOPEN='||-gzip -cdfq -- %s' less t.gz`

When you see the first screenful, go to another window and kill the `gzip` process that's waiting for `less` to read more input.  `less` won't notice or report the gzip failure.

Attached is a proposed patch to `less` that will cause it to diagnose input preprocessor failures like this. The idea is simple: if `pclose` returns a nonzero value, output an "Input preprocessor failed" diagnostic. Since this means the cleanup `edit((char*)NULL);` can output a diagnostic, this cleanup call needs to be before `quit`'s call to `deinit`. I'm by no means a `less` expert so perhaps you can think of a better fix.

[less-fix.txt](https://github.com/gwsw/less/files/8401663/less-fix.txt)

gwsw:
If I apply this patch, run your gzip test and then just type q to exit less (without killing the gzip process), I see the "Input preprocessor failed" message. This is on a Linux 5.12 system. It looks like pclose is returning 512, which I think means that gzip is returning an exit status of 2, which according to the gzip man page means "a warning occurred". I'm not sure exactly why that's happening, but it makes me think there is some risk of false error messages in this change.

gwsw:
A little more experimentation indicates that gzip returns 2 if its output is not fully consumed:
```
$ gzip -cdfq -- obj/t.gz | head -2
1
2
$ echo ${PIPESTATUS[0]}
```

eggert:
> A little more experimentation indicates that gzip returns 2 if its output is not fully consumed:

[...]

PR #315, opened 2022-12-26
Diagnose input preprocessor failures (#258)

eggert:
This pull request fixes issue #258 - it's like the patch proposed near the end of that issue, except updated so that it merges nicely with the master branch of 'less'.


main.c
```diff
@@ -409,6 +409,7 @@ quit(status)
 	else
 		save_status = status;
 	quitting = 1;
+	edit((char*)NULL);
```
gwsw, line 412: What is the purpose of moving the edit() call up here? I think this will break --redraw-on-quit, since the file will be closed when repaint() is called.

eggert: I moved the `edit` call up so that the diagnosis of the input preprocessor failure would appear in the prompt line on the user's screen. With the `edit` call where it is in `less` master, the file is closed and `less` moves on so the diagnostic message never appears on the user's screen.

Thanks for mentioning the `--redraw-on-quit` issue. To fix that, I updated the pull request to add a function `drain_altpipe` that diagnoses the input preprocessor failure without doing the rest of the `edit` call, and to call `drain_altpipe` at what I hope is the right point, leaving the call to `edit` alone. The updated pull request also merges from current master.

PR #317, opened 2022-12-28
Update function definitions to ANSI style

### LINKED ISSUE #316: 'less' should port to C23 by using function prototypes
One of the changes in the 2023 edition of the C standard is that support for K&R style function definitions has been removed. Since 'less' still uses these, it won't compile on a vanilla C23 compiler. And since nobody cares about pre-C89 compilers any more, the simplest fix is for 'less' to use C89-style function prototypes everywhere. It is 33 years after C89, after all.

GCC optionally warns about these obsolete usages; see attached build transcript.I expect some compilers will start erroring out entirely soon.

[lessbug.txt](https://github.com/gwsw/less/files/10302462/lessbug.txt)

davidschlachter:
Many of the function definitions in `less` are in the traditional (K&R) style. In this PR, I've updated them to the ANSI style. To do so, I used [`cproto`](https://invisible-island.net/cproto/cproto.html), fixed some complicated lines manually, and checked that nothing was missed by passing `-Werror=old-style-definition` to `gcc`. I also updated the function header generation scripts to expect the ANSI style. This would fix #316.

### DIFF
--- a/brac.c
+++ b/brac.c
@@ -23,12 +23,8 @@
  * The characters which serve as "open bracket" and 
  * "close bracket" are given.
  */
-	public void
-match_brac(obrac, cbrac, forwdir, n)
-	int obrac;
-	int cbrac;
-	int forwdir;
-	int n;
+	public void 
+match_brac(int obrac, int cbrac, int forwdir, int n)
 {
 	int c;
 	int nest;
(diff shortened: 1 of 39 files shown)


gwsw:
Fixed in 76a7805e1f0d794946ab7d8c6d60454bd3f244a5 (less-618).

Issue #320, opened 2023-01-01
Build issues with Autoconf 2.71


eggert:
Fedora 37, which uses Autoconf 2.71, complains mightily about `less` master; see below. I'll see if I come up with fixes. Fixes will likely assume Autoconf 2.53 (2002) or later, but that's safe now.

```
$ make -f Makefile.aut
mv -f ./help.c ./help.c.old
mv: cannot stat './help.c': No such file or directory
make: [Makefile.aut:59: help.c] Error 1 (ignored)
rm -rf help.c
./mkhelp.pl < less.hlp > help.c
if cmp -s help.c help.c.old; then mv -f help.c.old help.c; fi
mv -f ./funcs.h ./funcs.h.old
mv: cannot stat './funcs.h': No such file or directory
make: [Makefile.aut:68: funcs.h] Error 1 (ignored)
grep -h '^public [^;]*$' ./main.c ./screen.c ./brac.c ./ch.c ./charset.c ./cmdbuf.c ./command.c ./cvt.c ./decode.c ./edit.c ./filename.c ./forwback.c ./help.c ./ifile.c ./input.c ./jump.c ./line.c ./linenum.c ./lsystem.c ./mark.c ./optfunc.c ./option.c ./opttbl.c ./os.c ./output.c ./pattern.c ./position.c ./prompt.c ./search.c ./signal.c ./tags.c ./ttyin.c ./version.c ./xbuf.c | sed 's/$/;/' >./funcs.h
./mkutable -f2 Mn Me -- unicode/UnicodeData.txt > compose.uni
./mkutable -f2 Cf -- unicode/UnicodeData.txt > fmt.uni
./mkutable -f2 Cc Cs Co Zl Zp -- unicode/UnicodeData.txt > ubin.uni
cd .; autoheader; autoconf
configure.ac:14: warning: The macro `AC_CONFIG_HEADER' is obsolete.
configure.ac:14: You should run autoupdate.
./lib/autoconf/status.m4:719: AC_CONFIG_HEADER is expanded from...
configure.ac:14: the top level
configure.ac:18: warning: The macro `AC_ISC_POSIX' is obsolete.
configure.ac:18: You should run autoupdate.
./lib/autoconf/specific.m4:550: AC_ISC_POSIX is expanded from...
configure.ac:18: the top level
configure.ac:39: warning: The macro `AC_HEADER_STDC' is obsolete.
configure.ac:39: You should run autoupdate.
[...]

PR #321, opened 2023-01-01
Port to Autoconf 2.71 (#320)

eggert:
This consisted of running 'autoupdate' and then minimally fixing up to keep the patch small and the build working. The manual fixes were mostly just to fix indenting. However, one nontrivial fix was to stop using RETSIGTYPE which Autoconf 2.63 (2008) declared permanently obsolete. The only platform that cares about this is the OS-9 version for Microware C 3.2, and I expect the code will still compile and run anyway (perhaps with a warning), in the unlikely case that an OS-9 user builds with Makefile.o9c instead of Makefile.o9g or Makefile.o9u.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -9,13 +9,13 @@
 # contact the author, see the README file.
 
 # Autoconf initialization.
-AC_INIT(less, 1)
+AC_INIT([less],[1])
 AC_CONFIG_SRCDIR([forwback.c])
-AC_CONFIG_HEADER([defines.h])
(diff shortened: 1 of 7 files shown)


gwsw:
I'll look through this. However regarding RETSIGTYPE, I don't see any harm in leaving this a #define. I think it's useful to help identify which functions are signal handlers, and if autoconf doesn't want to support it anymore we can just define it manually.

gwsw:
Fixed in b85e10184efd13b3a7b192de35f80a6ad842ee57.

PR #355, opened 2023-03-25
powershell chcp 65001 but ConsoleOutputCP is 936

### LINKED ISSUE #37: Windows: UTF-8 input doesn't quite work
When building less on Windows (with Visual C++) utf-8 input (when the codepage for the current console is set to 65001), doesn't work correctly.

When I press / to enter search and then enter an "æ", for instance, it doesn't display anything and freezes up or starts flashing a lot and doesn't respond after that. On an older machine it doesn't do this, but instead turns the æ into "^@'" (and ø and å into "^@(" and "^@^Z" respectively).

The version shipped with git, for example, doesn't have this problem, but is also not built against Visual C++, but rather MSYS2.

zeromake:
fix #37

### DIFF
--- a/main.c
+++ b/main.c
@@ -71,8 +71,11 @@ extern int      first_time;
 int main(int argc, char *argv[])
 {
 #ifdef _WIN32
-	SetConsoleCP(65001);
-	SetConsoleOutputCP(65001);
+	UINT cp = GetConsoleCP();
+	UINT output_cp = GetConsoleOutputCP();
+	if (cp != output_cp) {
+		SetConsoleOutputCP(cp);
+	}
 #endif
 	IFILE ifile;
 	char *s;
--- a/main.c
+++ b/main.c
@@ -70,6 +70,10 @@ extern int      first_time;
  */
 int main(int argc, char *argv[])
 {
+#ifdef _WIN32
+	SetConsoleCP(65001);
+	SetConsoleOutputCP(65001);
+#endif
 	IFILE ifile;
 	char *s;
 
--- a/main.c
+++ b/main.c
@@ -71,10 +71,25 @@ extern int      first_time;
 int main(int argc, char *argv[])
 {
 #ifdef _WIN32
-	UINT cp = GetConsoleCP();
-	UINT output_cp = GetConsoleOutputCP();
-	if (cp != output_cp) {
-		SetConsoleOutputCP(cp);
+	char* less_charset = lgetenv("LESSCHARSET");
+	if (less_charset) {
+		UINT less_cp = 0;
+		// code page on https://learn.microsoft.com/en-us/windows/win32/intl/code-page-identifiers
+		if (!strcmpi(less_charset, "utf-8")) {
+			less_cp = 65001;
+		} else if (!strcmpi(less_charset, "gb2312")) {
+			less_cp = 936;
+		} else if (!strcmpi(less_charset, "shift_jis")) {
+			less_cp = 932;
+		}
+		if (less_cp != 0) {
+			if (less_cp != GetConsoleCP()) {
+				SetConsoleCP(less_cp);
+			}
+			if (less_cp != GetConsoleOutputCP()) {
+				SetConsoleOutputCP(less_cp);
+			}
+		}
 	}
 #endif
 	IFILE ifile;


gwsw:
I don't know enough about Windows to tell if this is reasonable, but it seems to me that the user should be able to select their code page, rather than having ```less``` assume that the code page is always UTF-8. Shouldn't it be possible for a user to use ```less``` with a code page other than UTF-8? @adoxa Ping to Jason.

adoxa:
The OP mentioned they were already using the UTF-8 codepage, so this code should not be doing anything (`chcp 65001` does what the code does). And yes, it will not be the code page the user expects and will totally break DBCS.

zeromake:
@adoxa 
chcp 65001 on cmd has effect, but on powershell no effect.
should it be the same as the LANG environment variable on Linux.

zeromake:
on cmd
``` cmd
>chcp 65001
Active code page: 65001
>.\less.exe .\utf8.txt
123abcæ测试アテイン
.\utf8.txt (END)
```
on powershell
```powershell
❯ chcp 65001
Active code page: 65001
❯ .\less.exe .\utf8.txt
123abc忙娴嬭瘯銈儐銈ゃ兂
.\utf8.txt (END)
```

zeromake:
try various acquisitions CP
``` c
printf("cp: %d %d %d %d\n", GetACP(), GetOEMCP(), GetConsoleCP(), GetConsoleOutputCP());
```
on cmd
```
cp: 936 936 65001 65001
```
on powershell
```
cp: 936 936 65001 936
```
Should be changed to
```c
#ifdef _WIN32
	UINT cp = GetConsoleCP();
	UINT output_cp = GetConsoleOutputCP();
	if (cp != output_cp) {
		SetConsoleOutputCP(cp);
	}
#endif
```

adoxa:
It seems PowerShell (7.2.4) will restore the original output CP after each command. However, that doesn't explain how you apparently started with 65001 and then switched back to 936. I'm still wary of this change, as it's *really annoying* when programs change my CP behind my back. Having said that, having the input & output pages different is unusual in itself, but which one is "right"? In this case it's input, but what if in other programs it's output?

Another approach would be to use `set LESSCHARSET=utf-8`, as that will use the wide-character output functions (when the output CP is not UTF-8).

zeromake:
@adoxa 
Do you mean using environment variables to specify the value of SetConsoleOutputCP?

adoxa:
Defining `LESSCHARSET=utf-8` puts less into UTF-8 mode; when the output code page is not also UTF-8, text is converted to wide and displayed that way, effectively bypassing the code page.

zeromake:
@adoxa 
I changed the code to something like this:
```c
#ifdef _WIN32
	char* less_charset = lgetenv("LESSCHARSET");
	if (less_charset) {
		UINT less_cp = 0;
		// code page on https://learn.microsoft.com/en-us/windows/win32/intl/code-page-identifiers
		if (!strcmpi(less_charset, "utf-8")) {
			less_cp = 65001;
		} else if (!strcmpi(less_charset, "gb2312")) {
			less_cp = 936;
		} else if (!strcmpi(less_charset, "shift_jis")) {
			less_cp = 932;
		}
		if (less_cp != 0) {
			if (less_cp != GetConsoleCP()) {
				SetConsoleCP(less_cp);
			}
			if (less_cp != GetConsoleOutputCP()) {
				SetConsoleOutputCP(less_cp);
			}
		}
	}
#endif
```

adoxa:
That's not really going to work, as less doesn't recognise those names. Besides, it would probably also be better done in the output, converting to wide. This is what I mean by setting utf-8 (less 631x, Win10.0.19044.1526, VS2019):

```
C:\Projects\less>chcp 936
Active code page: 936

C:\Projects\less>less -+c -LF nihongo.txt
鏃ユ湰瑾?

C:\Projects\less>set LESSCHARSET=utf-8

C:\Projects\less>less -+c -LF nihongo.txt
日本語

C:\Projects\less>chcp 65001
Active code page: 65001

C:\Projects\less>set LESSCHARSET=

C:\Projects\less>less -+c -LF nihongo.txt
日本語

C:\Projects\less>pwsh
PowerShell 7.2.4
Copyright (c) Microsoft Corporation.

https://aka.ms/powershell
Type 'help' to get help.

PS C:\Projects\less> less -+c -LF .\nihongo.txt
日本語
PS C:\Projects\less> exit

C:\Projects\less>chcp 936
Active code page: 936

C:\Projects\less>set LESSCHARSET=utf-8

C:\Projects\less>pwsh
PowerShell 7.2.4
Copyright (c) Microsoft Corporation.

https://aka.ms/powershell
Type 'help' to get help.

PS C:\Projects\less> less -+c -LF .\nihongo.txt
日本語
PS C:\Projects\less> exit
```

Issue #379, opened 2023-06-08
windows: mouse wheel scroll broken since v605


avih:
Specifically, mouse wheel scroll is broken on Windows since 7b977a2d ("Added support for UTF-8 commands on Windows (#261)") up to and including current master (v635).

@Konstantin-Glukhov since it's your commit, can you (or someone else) confirm that this commit broke mouse wheel? any idea how to fix it?

I debugged it a bit (at this exact commit), and this line:
https://github.com/gwsw/less/commit/7b977a2dbae980ac3c8b128a3c0e95adefc8e6bb#diff-1de596521a6bd2aaf938f5f87ef7be66407a9e4ee616f3f923574a9a1337a88cR2874
```c
x11mousebuf[2] = X11MOUSE_OFFSET + (((int)ip.Event.MouseEvent.dwButtonState < 0) ? X11MOUSE_WHEEL_DOWN : X11MOUSE_WHEEL_UP);
```
is still getting the negative/non-negative value correctly on scroll down/up, so it does not seem like an issue of incorrect `dwButtonState` value.

So I couldn't put my finger on anything which looks wrong, but it doesn't scroll. Scrolling the wheel also seems to mess with the prompt (the status line).

In the the previous commit ("open v604") it scrolls fine and the prompt is fine too.

I didn't try to follow it further.

Konstantin-Glukhov:
Could you be more specific which shell is affected and how it can be reproduced? I have no issue with the scroll using less version version 633 in Windows Terminal. Is it Dos shell C:\WINDOWS\system32\cmd.exe affected?

Konstantin-Glukhov:
I just tested less version 590 with both cmd.exe and Windows Terminal. I do not see any difference in mouse scroll behavior with version 633. Windows Terminal scroll works fine and cmd.exe scroll is disabled when less is launched. I could not notice anything happening to the prompt either - it stays the same in both cmd.exe and Windows Terminal.

avih:
Apologies if my description was not clear enough. Let me retry.

> I just tested less version 590 with both cmd.exe and Windows Terminal. I do not see any difference in mouse scroll behavior with version 633. Windows Terminal scroll works fine and cmd.exe scroll is disabled when less is launched

What you're describing is when `less` is launched without `--mouse`, where indeed the windows console doesn't scroll its content by default, while the windows terminal does scroll its content.

[...]

PR #381, opened 2023-06-10
Fix windows: mouse wheel scroll broken since v605 #379

Konstantin-Glukhov:
Separate mouse and keyboard events.

### DIFF
--- a/screen.c
+++ b/screen.c
@@ -137,6 +137,9 @@ extern int sc_height;
 
 #if MSDOS_COMPILER==WIN32C
 #define UTF8_MAX_LENGTH 4
+unsigned char utf8[UTF8_MAX_LENGTH];
+int utf8_size;
+int utf8_current_byte;
 struct keyRecord
 {
 	WCHAR unicode;
@@ -2859,11 +2862,22 @@ public int win32_kbhit(void)
 		ip.Event.KeyEvent.wVirtualKeyCode == VK_SHIFT ||
 		ip.Event.KeyEvent.wVirtualKeyCode == VK_CONTROL ||
 		ip.Event.KeyEvent.wVirtualKeyCode == VK_MENU);
-		
+
 	currentKey.unicode = ip.Event.KeyEvent.uChar.UnicodeChar;
 	currentKey.ascii = ip.Event.KeyEvent.uChar.AsciiChar;
 	currentKey.scan = ip.Event.KeyEvent.wVirtualScanCode;
 	keyCount = ip.Event.KeyEvent.wRepeatCount;
+
+    // If multibyte character, return its first byte
+    if (currentKey.ascii != currentKey.unicode) {
+        utf8_size = WideCharToMultiByte(CP_UTF8, 0, &currentKey.unicode, 1, &utf8, sizeof(utf8), NULL, NULL);
+        utf8_current_byte = 0;
+        if (utf8_size == 0 ) {
+            utf8_size == 1;
+            utf8[0] = '\0';
+        }
+        return (TRUE);
+    }
 
 	if (ip.Event.KeyEvent.dwControlKeyState & 
 		(LEFT_ALT_PRESSED | RIGHT_ALT_PRESSED))
@@ -2908,17 +2922,13 @@ public int win32_kbhit(void)
 public char WIN32getch(void)
 {
 	char ascii;
-	static unsigned char utf8[UTF8_MAX_LENGTH];
-	static int utf8_size = 0;
-	static int utf8_next_byte = 0;
 
 	// Return the rest of multibyte character from the prior call
-	if (utf8_next_byte < utf8_size)
+	if (utf8_current_byte < utf8_size)
 	{
-		ascii = utf8[utf8_next_byte++];
+		ascii = utf8[utf8_current_byte++];
 		return ascii;
 	}
-	utf8_size = 0;
 
 	if (pending_scancode)
 	{
@@ -2935,16 +2945,7 @@ public char WIN32getch(void)
 			continue;
 		}
 		keyCount --;
-		// If multibyte character, return its first byte
-		if (currentKey.ascii != currentKey.unicode)
-		{
-			utf8_size = WideCharToMultiByte(CP_UTF8, 0, &currentKey.unicode, 1, &utf8, sizeof(utf8), NULL, NULL);
-			if (utf8_size == 0 )
-				return '\0';
-			ascii = utf8[0];
-			utf8_next_byte = 1;
-		} else
-			ascii = currentKey.ascii;
+        ascii = currentKey.ascii;
 		/*
 		 * On PC's, the extended keys return a 2 byte sequence beginning 
 		 * with '00', so if the ascii code is 00, the next byte will be 
@@ -2992,4 +2993,3 @@ public void WIN32textout(char *text, int len)
 #endif
 }
 #endif
-


avih:
Thanks, I can confirm this fixes the mouse wheel scroll issue for me.

Nit: the new code is added with space-indenting, while the surrounding code uses tab indenting.


```c
	currentKey.unicode = ip.Event.KeyEvent.uChar.UnicodeChar;
	currentKey.ascii = ip.Event.KeyEvent.uChar.AsciiChar;
```
While this is not part of the current patch, and I guess it works, I believe C doesn't guarantee that `ip.Event.KeyEvent.uChar.AsciiChar` is the lower bits of `ip.Event.KeyEvent.uChar.UnicodeChar` (the value was set to union member `UnicodeChar` by `ReadConsoleInpoutW`, but the code reads it from the union member `AsciiChar`), so this would probably be a better assignment:
```c
currentKey.ascii = currentKey.unicode & 0xff;
```
(or even `& 0x7f`, because `char` is signed, so technically values above 127 would be implementation-defined behavior)

screen.c
```diff
@@ -2859,12 +2862,23 @@ public int win32_kbhit(void)
...
 	currentKey.scan = ip.Event.KeyEvent.wVirtualScanCode;
 	keyCount = ip.Event.KeyEvent.wRepeatCount;
 
+    // If multibyte character, return its first byte
+    if (currentKey.ascii != currentKey.unicode) {
+        utf8_size = WideCharToMultiByte(CP_UTF8, 0, &currentKey.unicode, 1, &utf8, sizeof(utf8), NULL, NULL);
+        utf8_current_byte = 0;
+        if (utf8_size == 0 ) {
+            utf8_size == 1;
+            utf8[0] = '\0';
```
avih, line 2877: As I've learned recently, if `WIN32getch` returns 0, then it's an indication that the next value which `WIN32getch` returns should be interpreted as a keyboard scan code rather than ascii/UTF8 byte.

So maybe `win32_kbhit` should just return FALSE if it can't convert it to UTF8?

Konstantin-Glukhov:
Yes, this "pending scan" logic is a little bit weird to me.
My fix is not working at all. It only works for scrolling a mouse wheel :-) I need to think more...
I also ran into a problem with Unicode surrogate pairs, though I think I almost found a solution.

avih:
> My fix is not working at all

Can you describe what the problem actually is? I.e. why mouse wheel ended got broken in that commit (or at the current code)?

avih:
> Yes, this "pending scan" logic is a little bit weird to me.

It's quite similar to how unicode works: ReadConsoleInput[W] delivers a value that doesn't quite fit into one byte, so WIN32getch splits it into two: 0 and the scan code, delivered one after the other in two consecutive calls.

The "buffer" is `currKey.scan` (one byte is enough), and the equivalent of `utf8_current_byte < utf8_size` is `pending_scancode`.

You can see that the unicode and the scancode are very similar at the begining of WIN32getch.

In fact, they might benefit from a shared infrastructure of queuing a sequence of return values. Currently unicode and scancode each have its own mechanism for basically the same thing.

Konstantin-Glukhov:
The fact that AsciiChar is signed char does not mean it is treated as a signed number. Which characters above 127 are UTF8 that are not correctly saved in AsciiChar? I want to test this theory.

avih:
> Which characters above 127 are UTF8 that are not correctly saved in AsciiChar?

What do you mean not correctly saved? What is not correctly saved?

> I want to test this theory.

What theory?

All the codepoints which are not ASCII-7 are above 127, for instance any non-English European or Asian character ends up as a sequence of UFT-8 bytes which are all bigger than 127.

I don't know what the context of the question is, the fact that it's stored in signed char doesn't necessarily mean it gets broken. But casting values between signed and unsigned, especially with different storage sizes, can sometimes break things.

avih:
> In fact, they might benefit from a shared infrastructure of queuing a sequence of return values. Currently unicode and scancode each have its own mechanism for basically the same thing.

Actually, `x11mousebuf` is also exactly the same thing - it translate a mouse event from ReadConsoleInput into a sequence of return values for `WIN32getch`.

So we have the mouse events which are queued at `x11mousebuf`, and the remaining length is at `x11mouseCount`.

We have unicode which is queued as UTF-8 at `utf8[UTF8_MAX_LENGTH];` and its length is maintained at `utf8_size - utf8_next_byte`.

And we have scan code which is buffered at `currKey.scan`, and its "length" is maintained at `pending_scancode`.

I really do think that a unified system of queuing a sequenc e of return values would make these three systems much much simpler.

Konstantin-Glukhov:
Maybe I misunderstood what you said about AsciiChar and UnicodeChar:

>  I believe C doesn't guarantee that ip.Event.KeyEvent.uChar.AsciiChar is the lower bits of ip.Event.KeyEvent.uChar.UnicodeChar (the value was set to union member UnicodeChar by ReadConsoleInpoutW, but the code reads it from the union member AsciiChar), so this would probably be a better assignment:
> 
> currentKey.ascii = currentKey.unicode & 0xff;

The way I interpret your statement is that AsciiChar != (UnicodeChar 0xFF). How is it possible if they are unioned in memory?

avih:
> Maybe I misunderstood what you said about AsciiChar and UnicodeChar:

Ah, now I understand :)

> The way I interpret your statement is that AsciiChar != (UnicodeChar 0xFF).

No. I said that C doesn't guarantee this. It might or might not be equal.

> How is it possible if they are unioned in memory?

If the architecture is Big Endian, then `AsciiChar` will probably be the most significant byte of `UnicodeChar`, but if it's Little Endian, then it will probably be the lower byte.

I don't think there's a Windows system which is Big Endian, so I don't think it's an issue.

But what is the purpose of that test? to check whether it exceeds 8 bits in value? then you could do, for instance:

```c
if (UnidoceChar != (unsigned char)UnicodeChar)
    ...
```
or something with `& ~0x7f` or whatever.

Accessing a union member which was not set is a bit of a gray area.

avih:
If you want to test whether it will be more than one UTF-8 byte, then just check whether `UnicodeChar` is bigger than 127.

Konstantin-Glukhov:
The reason the mouse wheel is broken because of the UTF code in WIN32getch() overwrites the value of currentKey.ascii variable. The UTF code needs to be in win32_kbhit() not to step on the mouse code.

I am puzzled why the mouse code is inside the loop and the key events code is outside the loop. Shouldn't they follow similar logic?

Also why are there two variables for the mouse buffer: x11mousePos and x11mouseCount? Isn't one enough to move around the buffer? Couldn't it be like:

```
if (x11mousePos < 5)
	{
		currentKey.ascii = x11mousebuf[x11mousePos++];
		keyCount = 1;
		return (TRUE);
	}

```

avih:
> The UTF code needs to be in win32_kbhit()

I agree.

> I am puzzled why the mouse code is inside the loop and the key events code is outside the loop. Shouldn't they follow similar logic?

They do follow similar logic: if the code knows right away that it's going to ignore this INPUT_RECORD, then it will continue the loop. This happens with mouse move events, KEY_EVENT which is not "down", key events without either scan code or unicode, etc. It's not only for mouse.

> Also why are there two variables for the mouse buffer: x11mousePos and x11mouseCount? Isn't one enough to move around the buffer? Couldn't it be like:

There's more than one way to write the same code. Your example will probably work fine too. It comes down to personal preference, and at this case I'm guessing the author preferred to keep it more generic, for instance if one day the mouse sequence could have different lengths.

Konstantin-Glukhov:
In which case wVirtualScanCode can be 0?

avih:
So I unified the unicode/scancode/x11mouse keyboard queues. All three need to return more than one value via `WIN32getch()`, and each maintained its own variable, queue, and system.

Now they all use the same queue.

Also moved the unicode code from `WIN32getch` to `win32_kbhit` where all the other code which converts `ReadConsoleInput` into one or more return values is.

This happens to fix the mouse wheel issue without any specific effort.

Also fixes repeat count (at the KEY_EVENT) which was not working for unicode.

The Unicode still doesn't handle surrogate, ans uses the exact same "system" as before: read one `WCHAR` and try to convert it to UTF-8.

The patch is the top commit here https://github.com/avih/less/commits/winkbq

For now I'm not opening a PR for it, but leaving it there if others want to have a look.

Konstantin-Glukhov:
Is mouse supposed to do text selection and copy/paste?

Konstantin-Glukhov:
I opened a new pull request #385 with the working code

PR #383, opened 2023-06-13
Windows/MS-DOS: decouple SGR state from screen colors

### LINKED ISSUE #377: windows: SGR-1 incorrectly resets the color
If invoking `less -R` and there's an SGR sequence of some color, followed by SGR sequence of bold/bright, then the color is reset and only bright is displayed.

The other way around (bright, then some color) does work correctly, i.e. it shows the color in bright.

For instance this (using [busybox-w32](https://github.com/rmyorston/busybox-w32) for `printf`):
```sh
busybox printf "\033[32m green \033[1m bright \033[m reset \033[1m bright \033[32m green \033[0m\n" | .\less.exe -RXE
```

prints SGR32 green SGR1 bright SGR0 reset SGR1 bright SGR32 green, which should display as normal-green, bright-green, plain-white, bright-white, bright-green.

However, in practice the second text is displayed as bright-only (bright cyan actually, as that's the default bold color of `less` on Windows), instead of bright green.

Notice that the other way around (the last "green" which is after SGR1 followed by SGR32) is correctly bright-green.

Tested with latest official release for windows ([v633](https://github.com/jftuga/less-Windows/releases/tag/less-v633)) but also with my own builds, and this issue exists since at least v540 (in 2018).

Screenshot in `cmd.exe` window:
less-color-bright-win32

EDIT: interestingly, green followed by bright at the same SGR sequence does show correctly, e.g. `printf "\033[32;1m bright-green \033[m" | less -R` does show as bright-green.

avih:
The trigger for this commit was to make the -D<x> mapping more predictable. Instead of modifying win_flush (~250 LOC), the SGR and colors logic was refactored out of win_flush, which was an opportunity to make the logic simpler and more rebust.

See the commit message for full description of the changes.

Fixes #377

### DIFF
--- a/output.c
+++ b/output.c
@@ -84,6 +84,153 @@ static char *ob = obuf;
 static int outfd = 2; /* stderr */
 
 #if MSDOS_COMPILER==WIN32C || MSDOS_COMPILER==BORLANDC || MSDOS_COMPILER==DJGPPC
+
+typedef unsigned t_attr;
+
+#define A_BOLD      (1u<<0)
+#define A_ITALIC    (1u<<1)
+#define A_UNDERLINE (1u<<2)
+#define A_BLINK     (1u<<3)
+#define A_INVERSE   (1u<<4)
+#define A_CONCEAL   (1u<<5)
+
+/* long is guaranteed 32 bits, and we reserve bits for type + RGB */
+typedef unsigned long t_color;
+
+#define T_DEFAULT   0ul
+#define T_ANSI      1ul  /* colors 0-7 */
+
+#define CGET_ANSI(c) ((c) & 0x7)
+
+#define C_DEFAULT    (T_DEFAULT <<24) /* 0 */
+#define C_ANSI(c)   ((T_ANSI    <<24) | (c))
+
+/* attr/fg/bg/all 0 is the default attr/fg/bg/all, respectively */
+typedef struct t_sgr {
+	t_attr attr;
+	t_color fg;
+	t_color bg;
+} t_sgr;
+
+static const t_sgr SGR_DEFAULT; /* = {0} */
+
+static void update_sgr(t_sgr *sgr, unsigned char code)
+{
+	switch (code)
+	{
+	case  0: *sgr = SGR_DEFAULT; break;
+
+	case  1: sgr->attr |=  A_BOLD; break;
+	case 22: sgr->attr &= ~A_BOLD; break;
+
+	case  3: sgr->attr |=  A_ITALIC; break;
+	case 23: sgr->attr &= ~A_ITALIC; break;
+
+	case  4: sgr->attr |=  A_UNDERLINE; break;
+	case 24: sgr->attr &= ~A_UNDERLINE; break;
+
+	case  6: /* fast-blink, fallthrough */
+	case  5: sgr->attr |=  A_BLINK; break;
+	case 25: sgr->attr &= ~A_BLINK; break;
+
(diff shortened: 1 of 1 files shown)


avih:
It's been two months since I opened this PR, I'd appreciate a comment.

gwsw:
I do plan to merge this. I have been working on a bug-fix release for the past two months which has gone through several beta releases and has just been released for general use yesterday. Assuming I don't get any showstopper bug reports in the next week or so, I will resume new development after that and this will be one of the first issues I look at.

gwsw:
Is there a reason to require the -Da option to enable sgr_mode? Why don't we automatically enable sgr_mode when we set vt_enabled (that is, when SetConsoleMode(ENABLE_VIRTUAL_TERMINAL_PROCESSING) succeeds)? I see there was some discussion of this in #377 but I'm not sure I fully understand the reasoning. If there's a reason to want non-VT behavior on a VT terminal, maybe the current -Da behavior should be the default and a new option could disable it?
Ping @adoxa

avih:
> Why don't we automatically enable sgr_mode when we set vt_enabled

Generally speaking, there shouldn't be a reason if we presume that the terminal will do a better job in typical use cases than the emulation in `less`, and that would probably be a fair assumption. The VT mode is quite capable.

We do know that VT mode supports true colors - which the emulation doesn't, and its state management is probably much better than the emulation, at least before this PR, but probably also after (for instance, the PR still doesn't ignore correctly the values of 256/true colors. I do have an early version that not only doesn't ignore them, it actually converts them to 4 bit color, but I don't intend to PR it, it was more of an experiment, which went well - see [here](https://github.com/avih/less/commit/fe285efd21963a54be5d0a136f925f9e393ab02a), and this PR is a thinned-down and cleanned-up version of that experiment).

But for me, there are few reasons:
1. This PR is about improving the non "sgr mode"  behaviour, so I didn't think that changing the default mode belongs here.
2. Personally, I like that bold-alone (SGR 1) also use a brighter color, rather than only emboldened font/glyph, but the default with the Windows terminal is to use only emboldened font, and I dislike it. The emulation uses the high intensity bit for bold, so for my taste it's much nicer.
3. I also think that Underline doesn't look very good with the windows terminal (it uses 1px line, which is too thin for my taste, especially with bigger font size), but the emulation in `less` can use something which looks better for my taste.

So because bold and underline happen frequently in my use cases (syntax highlight, etc), I prefer the emulation -Dx values.

That being said, I also changed the defaults -Dx values to be more neutral, [see here](https://github.com/avih/less/commit/982b3275401a043897c49fe4798bee80d981e878), so my -Dx colors are a bit different than less-master. Feel free to take/copy this commit and push it if you like it more than the current default.

Overall, -Da probably should be the default, but personally I prefer the emulation currently because bold and underline look better for my taste than in VT mode.

It should be noted that the windows terminal can actually support brighten+embolden for SGR 1, but it's undocumented and non intuitive - one needs to set the default foreground color to be identical to the while (SGR 7) color, and then SGR 1 also chooses the bright color of SGR 7 (i.e. color 15 in x256 colors). So there's that.

adoxa:
It's about legacy as much as anything - I started with DOS, where colours were the only option for visible attributes. Since man pages switched to sequences, that's what I kept. I very rarely view ANSI files, but when I did that's what -Da is for. I still prefer the colours to the attributes, but since that's more or less supported directly now, anyway, -Da is not really relevant anymore (turn the colours off, rather than turning sequences on).

PR #384, opened 2023-06-14
Windows: fix iread

### LINKED ISSUE #378: windows: unexpected abort on repeated page-down with color content
Tested `less` versions: latest official windows build (v633), but also my own builds since v570 (specifically, since 26e91bd "Allow ^X to exit F mode on Windows") upto and including v635.

Tested windows versions: Windows 10 and Windows 7.

Steps to reproduce:
- Set Windows autorepeat to a high value (if win10 - via the classic control panel). I set it to the maximum, which is about 30 repeats/sec.
- Maximize the console window (`cmd.exe`).
- Invoke `less -R file` where `file` is a big file with color escape sequences.
- Hold down page-down so that it auto-repeats and pages continuously.

Expected result:
It pages till the end of the file.

Actual result:
`less` aborts sooner rather than later. For me that's typically after less than 10 pages.

Here's a sample C file which was highlighted using GNU source-highlite, and reproduces the issue for me: 
[tccgen.c.zip](https://github.com/gwsw/less/files/11679789/tccgen.c.zip)

Additional info:

The offending code seems to be this at `os.c`, which was originally added at 26e91bdf:
```c
#if MSDOS_COMPILER==WIN32C
        if (win32_kbhit() && WIN32getch() == intr_char)
        {
                sigs |= S_INTERRUPT;
                reading = 0;
                return (READ_INTR);
        }
#endif
```
If I disable this code completely, e.g. by replacing `WIN32C` with `WIN32C_XXXBUG`, then the issue goes away.

When not disabled, and when it aborts, `win32_kbhit()` is true, and `WIN32getch()` returns `0`, which apparently is not `intr_char` (which seem to be 24 by default), so the code block of `sigs |= S_INTERRUPT;` etc is NOT entered, and yet `less` aborts.

The issue is more likely to happen with a big console window.

I'm guessing that the cause is that `less` renders slower than requested (high PGDN input rate, big console, slow-to-render colors), and something gets messed up, but I didn't dig deep enough to figure out what.

Another option is that it needs `ungetc()` to revert the `WIN32getch()` call, but I couldn't figure out how (i.e. which function to use to unget it). But also, `WIN32getch()` returning 0 also seems a bit unexpected...

adoxa:
The Windows version of iread removed the character, causing an extended sequence to be incorrectly split.  Add an unget function to put the character back.  Resolves #378.

### DIFF
--- a/os.c
+++ b/os.c
@@ -244,11 +244,18 @@ start:
 	}
 #else
 #if MSDOS_COMPILER==WIN32C
-	if (win32_kbhit() && WIN32getch() == intr_char)
+	if (win32_kbhit())
 	{
-		sigs |= S_INTERRUPT;
-		reading = 0;
-		return (READ_INTR);
+		int c;
+
+		c = WIN32getch();
+		if (c == intr_char)
+		{
+			sigs |= S_INTERRUPT;
+			reading = 0;
+			return (READ_INTR);
+		}
+		WIN32ungetch(c);
 	}
 #endif
 #endif
--- a/screen.c
+++ b/screen.c
@@ -2956,6 +2956,16 @@ public char WIN32getch(void)
 
 	return ascii;
 }
+
+/*
+ * Restore the character read during iread.
+ */
+public void WIN32ungetch(int ch)
+{
+	currentKey.unicode = ch;
+	++keyCount;
+	pending_scancode = 0;
+}
 #endif
 
 #if MSDOS_COMPILER


avih:
Thanks. I can confirm it fixes #378 for me.

But I don't think it would work for unicode chars, because the unicode path checks `utf8_current_byte < utf8_size` before it looks at `currentKey.unicode`.

Also note that #381 wants to change a bit how unicode is handled.

avih:
> I don't think Unicode'll be an issue, since it's really only the combination of page down during reading the file that has the problem

Yes, this fixes the issue where it needs to unget the `0` which was the "prefix" of a scan code.

However, it doesn't necessarily work correctly in all other cases.

For instance, it unconditionally increases `keyCount`, so obviously any case where `WIN32getch` didn't decrease it - will not be restored to exactly the same state as before.

Cases where `WIN32getch` doesn't change `keyCount`:
- Any char of a UTF-8 sequence except the first.
- The scan code which follows an `0` (right after `return ((char)(currentKey.scan & 0x00FF));`).
- `\003` in case of `ABORT_SIGS()`.

The main underlaying issue is that the state is non trivial and composed of many variables: x11mouse state, scan code state, utf8 state, each with its own variables and buffers.

So at the very least it's very cumbersome to tell exactly what the previous state was by looking at the current state and the last return value from `WIN32getch`, and it's most probably not worth trying.

It's OK for now because it can successfully unget the `0`-prefix of a scan code, which was the main trigger to add `WIN32ungetch`, but I intend to submit a PR to make the keyboard state much simpler, and consequently make `WIN32ungetch` work correctly in all cases.

So just an FYI.

avih:
I've submitted #388 which should make `WIN32ungetch` work on 100% of the cases.

I opted for a much smaller fix, without refactoring the KB queue system.

If anyone is interested, a proof of concept patch which unifies the KB queues and also allows a trivial `WIN32ungetch` is available at the top commit in this branch https://github.com/avih/less/commits/winkbq

It was written before `WIN32ungetch` was added, so it's not implemented, but it would be:
```c
public void WIN32ungetch(int ch)
{
	if (kbq_free())
		kbq_ungetc(c);
}
```

I don't intend to open a PR for it because the KB ququq system does work for the most part, even if the state can be non trivial, but I can open a PR if @gwsw shows explicit interest.

Issue #368, opened 2023-05-20
Less fails to read pipe from sudo if sudo prompts for a password


rpigott:
The following reproduces the issue somewhat reliably in both zsh and bash when sudo is configured to promprt for a password:

`sudo -k seq 100 | less`

We expect less to display the integers 1..100 but attempting to enter the sudo passphrase usually results in getting stuck "Waiting for data...".

This is a regression. I bisected it to 953257f9030d:

<details>
<summary>git bisect log</summary>

```
git bisect start
# status: waiting for both good and bad commits
# good: [22e4af5cccbfab633000c7d610f868a868ad6e1a] v608
git bisect good 22e4af5cccbfab633000c7d610f868a868ad6e1a
# status: waiting for bad commit, 1 good commit known
# bad: [0d7ca880375bde651313f03f7115d26921b43722] v633
git bisect bad 0d7ca880375bde651313f03f7115d26921b43722
# bad: [fb09748090dfa58de138978a8a06b8acafc48727] Cleanup
git bisect bad fb09748090dfa58de138978a8a06b8acafc48727
# bad: [fb09748090dfa58de138978a8a06b8acafc48727] Cleanup
git bisect bad fb09748090dfa58de138978a8a06b8acafc48727
# good: [ece2600cb0f0c8339cfe00ffa8e2dbc28df38622] v609
git bisect good ece2600cb0f0c8339cfe00ffa8e2dbc28df38622
# bad: [a03f8263a009b5e60aff50e438f08faf7daaa3a2] Open v613.
git bisect bad a03f8263a009b5e60aff50e438f08faf7daaa3a2
# bad: [1216c2baa3ccbafe0367edf06209e62d170536ff] Files genrated from updated Unicode data.
git bisect bad 1216c2baa3ccbafe0367edf06209e62d170536ff
# good: [ab5ad29445368e7da91a80d433a3f88e42be6127] Merge branch 'master' of https://github.com/gwsw/less
git bisect good ab5ad29445368e7da91a80d433a3f88e42be6127
# bad: [112b028f1421f093e9f79600fb67e6a78ccdd414] Add unicode-check target.
git bisect bad 112b028f1421f093e9f79600fb67e6a78ccdd414
# bad: [8e3ec369acfa1741c2e84a1655914c40c63a3f47] More reasonable number.
git bisect bad 8e3ec369acfa1741c2e84a1655914c40c63a3f47
[...]

PR #385, opened 2023-06-15
Update screen.c to fix broken mouse wheel by UTF code

Konstantin-Glukhov:
This is a new pull request instead of #381

### DIFF
--- a/version.c
+++ b/version.c
@@ -985,7 +985,7 @@ v633  5/3/23    Fix build on systems with ncurses/termcap.h or ncursesw/termcap.
 v634  5/29/23   Allow program piping into less to access tty;
                 fix search modifier ^E after ^W.
 v635  6/2/23    Fix crash with ! search modifier.
-v636            Fix -D in MS-DOS build; fix mouse wheel in MS-DOS build.
+v636  6/18/23   Fix -D in MS-DOS build; fix mouse wheel in MS-DOS build.
 */
 
-char version[] = "636x";
+char version[] = "636";


avih:
This PR and single commit tries to solve two completely unrelated problems:
- Mouse wheel scroll when using `--mouse` is broken since UTF support was added.
- UTF support doesn't work for surrogate pairs, like emojis.

It would be best to at least separate the two changes into two commits, IMHO, or just leave the surrogate pairs change for another time or PR.

(I regret ever mentioning the surrogate pairs issue. It's not critical, and probably not worth the complexity added by solving it).

Specifically, the surrogate pairs change can now make `win32_kbhit` a liar, because eventhough `win32_kbhit` returns true (which means that `WIN32getch` can be called withouit blocking), `WIN32getch` can now block forever - if the second part of the pair doesn't arrive.

`ReadConsoleInputW` is blocking, and it makes no guarantee that if the first half of a surrogate pair was delivered, the second half is already queued too, or that it's going to arrive at any time at all. So the code can't make such assumption.

avih:
@Konstantin-Glukhov @gwsw 

As I mentioned here https://github.com/gwsw/less/pull/381#issuecomment-1592111635 I have a pending change to unify the "return a sequence of values" code between the x11 mouse code (6 return values), scancode (2 return values), and unicode (1-4 return values).

My change happens to solve the mouse wheel scroll issue simply because it uses a single queue instead of 3 different queue systems, which was a bit buggy.

Also, it does NOT try to solve the surrogate pairs problem. It's so exactly like before - it reads one WCHAR at a time and tries to converts it on its own to UTF-8.

I didn't open a PR because I didn't want to add a "competing" PR, but would you two mind if we try to merge that unified queue first, and then @Konstantin-Glukhov can try to solve the surrogate pairs issue independently?

Konstantin-Glukhov:
Actually I mentioned surrogate pair problem myself -:). I do not mind merging our codes without surrogate pair handling. Though your concern can be easily addressed by first using PeekConsoleInputW  to make sure the low surrogate is present, and if not return FALSE. Also your code needs to take the while() condition from mine for UTF8 to be working.

avih:
For future reference, I explained exactly why all mouse-related things were broken since Unicode support was added, here https://github.com/gwsw/less/issues/379#issuecomment-1592677290 .

avih:
> would you two mind if we try to merge that unified queue first

> I do not mind merging our codes without surrogate pair handling

I'm not sure if I was clear.

I meant to merge my unified-queue patch into master first, and then @Konstantin-Glukhov can send a new PR for only the surrogate pairs fix, if he wants.

My unified queue patch happens to fix the mouse issues only as a side effect, because with a unified queue there's no more clash between the x11mouse queue and the unicode code.

gwsw:
The mouse issue should be fixed in 6fd6739c00901764ceb6dea2c681b0e000d5028c.
I guess the only remaining issue that this PR addresses is surrogate pairs. The current release which is now in beta is intended to be a small bug-fix release without new and/or risky features. Does anyone think that anything in this PR should be included in the current beta?
@avih @Konstantin-Glukhov @adoxa

adoxa:
No, it can wait.

avih:
Surrogate pairs never worked and no one ever reported it. IMHO it's certainly not urgent to fix it....

avih:
As for the surrogates pair code, did you check that it actually works and `less` can use such input?

For instance by adding this emoji to some text file: 😀  (U+1F600)

Then viewing this text file in `less`, then search this emoji by typing `/` and pasting it and ENTER, and then `less` finds it and highlights it at the text file?

I'm asking because I see that the current code seem to depend on two things which I empirically observed to be untrue:
- That surrogate pair halfs are delivered with `ip.Event.KeyEvent.bKeyDown` (in my experience they never arrive with `bKeyDown`).
- The second (low) half arrives immediately after the first (high) half (I observed several input events between them, so I'd guess that immediately reading another event will not get the second half).

Konstantin-Glukhov:
The keyboard input doesn't work yet, pasting also does not work. The command line doesn't display the surrogate pair correctly. It has something to do with the way code is sent to the console. I didn't have time to look into this. I never noticed events in between, unless they are control keys, these do not change the content ov the Unicode field. One thing I can confirm that works is retrieving an emoji from the search history and searching for it.

As you said we need to postpone surrogate pairs fix for later.

Konstantin-Glukhov:
Should I cancel this PR and you submit your own?

avih:
> I never noticed events in between, unless they are control keys, these do not change the content ov the Unicode field

Yes, I also observed control keys, however, the loop just above your code apparently exits right when the first half is read, and then the surrogate pair code immediately reads another input - and the loop doesn't get a chance to skip those control keys (or a key-up event, if there is one), so the surrogate code will want to read the second half, but it would get a different thing instead.

At least that's how it seems to me by looking at the code. I could be wrong, which is why I asked if it works.

> Should I cancel this PR and you submit your own?

My observation about how unicode input works at the console were in some other project.

I'm not working on this for `less`, I don't have a PR for this, and currentIy I don't intend to work on this (and personally I think it would be more complex than fun or useful).

You should do what's best for you. I can't tell you what to do with your time.

avih:
> The current release which is now in beta is intended to be a small bug-fix release without new and/or risky features.

Is the release process described someplace? If not, could you please describve it here?

Specifically, did this refer to v636 as a "small bugfix release beta"?

How should one know whether the next release is "small bugfix beta", or anything else, before it was released? For instance, I don't see it mentioned at the "open vXXX" commit messages...

I'm asking because the web page lists v635 as the latest beta, eventhough v636 was tagged at the git repo a week ago, and currently it's already in the v637 cycle.

I'm also asking because I have two open PRs which did not recieve even a comment in weeks. Initially I thought it's due to this "small bugfix release, I'll handle bigger things later", but this has not happen with my PRs that I can tell.... So it would be nice to get even something like "thanks, I hope to get to this next release/week/month/year" to let me know where it stands...

gwsw:
This part of the release process isn't documented. less-633 was a major release and was released about two months ago. Since then a few bugs in it have been found, of which #368 was probably the most important. I want to get these fixed and in a new stable release before adding any new features. Any issues that have been filed but are too large or risky for this stabilization release will be handled later. I can't say exactly when that will be, but a normal (not bugfix) release usually takes a several months to get through the beta cycle.

avih:
Thanks fo the info.

So in general, the cycle is roughly a release every few months, with intermediate tags which serve as development checkpoints/testing/beta, possibly tagged after adding meaningful changes, and if there were some outstanding issues in a release, then the next one might be a small bugfix release, which would likely have a shorter cycle, but still with some intermediate beta tags?

And in this case, 633 had mainly one outstanding issue, and so 634, 635, 636....(estimate how many are likely to be?) serve as beta/intermediate tags before the next small bugfix release - after which you'll start taking more risky changes?

Maybe it's worth noting at the commit message  of "Opening vXXX" what's the intent of XXX, especially if it's limited in some way compared to more "standard" intermediate development tag (for instance as was case with all the tags after 633 so far)?

gwsw:
Yes that's all correct.

It's often not possible to predict the usage of a release at the time I build it. For example, when I built less-635 I thought it was going to be the next beta release, but some issues that arose after it was built have changed that plan. The same thing happened for less-636. I now believe that less-637, which I will build soon, will be the next beta, but whether that will be true depends on future events.

avih:
A suggestion, if I may.

As far as I can tell, currently, a PR may idle without any comment, possibly for months, because the next release is supposed to be a small bugfix release without risky new code, which is understandable.

However, meanwhile, the author may forget important details about it. It clearly wouldn't be fresh in their mind as it was right after they authored it.

How about reviewing it normally in such case, soon after the PR is submitted - if time permits, let the author address review comments as needed, discuss stuff which needs resolving, rinse and repeat until everyone is happy, and _then_, once it's reasonably complete, then let it idle till the time comes to merge it?

It could also give chance to address bigger issues early, such as undesirable approach, etc.

Surely, there could pass some time between the review and the merge, and so it might need a rebase, or possibly even become irrelevant altogether

However, that would be exactly the same also if the review process starts only when the maintainers are ready to merge more risky code, so even if there could be cases where it could save some time to start the review late, I'd guess that typically it would be just fine, with the big advantage that issues are addressed early while the code is fresh for the developer.

What say you?

PR #388, opened 2023-06-17
windows: improve WIN32ungetch

avih:
WIN32ungetch does serve the purpose it was added for - unget a zero value which is the first-of-two return values of a scan code.

However, there were some cases which never worked, like ungetting any UTF-8 byte except the first (of a sequence of up to 4), where the next WIN32getch call simply ignores the ch argument to WIN32ungetch.

Additionally, some other cases got broken at 6fd6739c , because the unget code was depending on buggy unicode behavior - which is now fixed. E.g. unget the second-of-two scan code return values is broken.

Instead of trying to modify the quite-complex KB queue state (unicode state, x11 mouse state, scan code state) in a way which will restore the current state after the next WIN32getch call, this commit simply bypasses the KB queue state, and should ensure it always works.

### DIFF
--- a/screen.c
+++ b/screen.c
@@ -2902,8 +2902,17 @@ public int win32_kbhit(void)
 	return (TRUE);
 }
 
+static int unget_pending, unget_data;  /* = 0 */
+
 /*
  * Read a character from the keyboard.
+ *
+ * Known issues:
+ * - WIN32getch API should be int like libc (with unsigned char values or -1).
+ * - The unicode code below can return 0 - incorrectly indicating scan code.
+ * - UTF16-LE surrogate pairs don't work (and return 0).
+ * - If win32_kbhit returns true then WIN32getch should never block, but it
+ *   will block till the next keypress if it's numlock/capslock scan code.
  */
 public char WIN32getch(void)
 {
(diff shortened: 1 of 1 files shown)


avih:
Thanks for merging it.

I noticed that the commit was merged with the author `avih <avih@users.noreply.github.com>`, but this PR has a more detailed author - https://patch-diff.githubusercontent.com/raw/gwsw/less/pull/388.patch

Did github change the author during the merge? I don't recall it happening in the past with PRs...

gwsw:
I don't know why the Author name is set like that. I just clicked "Squash and merge" like I always do to merge a PR.
I have added the check in win32_kbhit as you suggest.

avih:
> I don't know why the Author name is set like that. I just clicked "Squash and merge"

Hmm, I'll try to figure out why that happened.

> I have added the check in win32_kbhit as you suggest.

Yeah, I noticed. Thanks.

avih:
> I don't know why the Author name is set like that. I just clicked "Squash and merge" like I always do to merge a PR.

Right. I've been told it might have been the "squash" part.

You could have also done "Rebase and Merge", which should be effectively the same as "rebase and squash" if there's only one commit - as was the case here, though possibly without github rewriting the author.

If the PR has more than one commit then "rebase and merge" would also preserve the individual commits, if you want that.

For what it's worth, In other projects when I merge a PR I typically use "rebase and merge" (and if it needs to be squashed first, then I'd typically ask the author to do that, if they know how).

So just FYI, and maybe next time try "Rebase and merge" ;)

PR #390, opened 2023-06-20
mouse: bind right-click as jump-to-"#"-mark

avih:
Previously any mouse-button click set the "#" mark.

With this PR, left-click sets the "#" mark, and right-click jumps to the "#" mark.

There are two commits: the first fixes mouse-button-down encoding on windows in preparation, and the second adds the right-click binding as jump-to-mark.

### DIFF
--- a/decode.c
+++ b/decode.c
@@ -439,9 +439,9 @@ static int mouse_wheel_up(void)
 }
 
 /*
- * Return action for a mouse button release event.
+ * Return action for the left mouse button trigger.
  */
-static int mouse_button_rel(int x, int y)
+static int mouse_button_left(int x, int y)
 {
 	/*
 	 * {{ It would be better to return an action and then do this 
@@ -450,6 +450,23 @@ static int mouse_button_rel(int x, int y)
 	if (y < sc_height-1)
 	{
 		setmark('#', y);
+		screen_trashed();
+	}
+	return (A_NOACTION);
+}
+
+/*
+ * Return action for the right mouse button trigger.
+ */
+static int mouse_button_right(int x, int y)
+{
+	/*
+	 * {{ unlike mouse_button_left, we could return an action,
+	 *    but keep it near mouse_button_left for readability. }}
+	 */
+	if (y < sc_height-1)
+	{
+		gomark('#');
 		screen_trashed();
 	}
 	return (A_NOACTION);
@@ -484,6 +501,7 @@ static int getcc_int(char *pterm)
  */
 static int x11mouse_action(int skip)
 {
+	static int prev_b = X11MOUSE_BUTTON_REL;
 	int b = getcc() - X11MOUSE_OFFSET;
 	int x = getcc() - X11MOUSE_OFFSET-1;
 	int y = getcc() - X11MOUSE_OFFSET-1;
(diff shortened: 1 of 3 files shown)


avih:
> Alternatively, and possibly even simpler and better, handle the buttons state at `x11mouse_action` so that click is reported on button-up on any system which uses the legacy method

So, for reference, that's the legacy mouse handler patch to add support for specific buttons, which is currently in this PR, and it triggers on button-down - which is OK, but suboptimal:
```diff
static int x11mouse_action(int skip)
 {
        int b = getcc() - X11MOUSE_OFFSET;
        int x = getcc() - X11MOUSE_OFFSET-1;
        int y = getcc() - X11MOUSE_OFFSET-1;
        if (skip)
                return (A_NOACTION);
        switch (b) {
        default:
                return (A_NOACTION);
        case X11MOUSE_WHEEL_DOWN:
                return mouse_wheel_down();
        case X11MOUSE_WHEEL_UP:
                return mouse_wheel_up();
-       case X11MOUSE_BUTTON_REL:
-               return mouse_button_rel(x, y);
+       case X11MOUSE_BUTTON1:
+               return mouse_button_left(x, y);
+       /* is BUTTON2 the rightmost button with 2-buttons mouse? */
+       case X11MOUSE_BUTTON2:
+       case X11MOUSE_BUTTON3:
+               return mouse_button_right(x, y);
        }
 }
```

And that's an alternative patch which adds specific-buttons support, but triggers on button-up - which is nicer (and would make the windows switch to SGR reporting redundant):

```diff
 static int x11mouse_action(int skip)
 {
+       static int prev_b = X11MOUSE_BUTTON_REL;
        int b = getcc() - X11MOUSE_OFFSET;
        int x = getcc() - X11MOUSE_OFFSET-1;
        int y = getcc() - X11MOUSE_OFFSET-1;
        if (skip)
                return (A_NOACTION);
        switch (b) {
        default:
+               prev_b = b;
                return (A_NOACTION);
        case X11MOUSE_WHEEL_DOWN:
                return mouse_wheel_down();
        case X11MOUSE_WHEEL_UP:
                return mouse_wheel_up();
        case X11MOUSE_BUTTON_REL:
-               return mouse_button_rel(x, y);
+               /* to trigger on button-up, we check the last button-down */
+               switch (prev_b) {
+               case X11MOUSE_BUTTON1:
+                       return mouse_button_left(x, y);
+               /* is BUTTON2 the rightmost button with 2-buttons mouse? */
+               case X11MOUSE_BUTTON2:
+               case X11MOUSE_BUTTON3:
+                       return mouse_button_right(x, y);
+               }
+               return (A_NOACTION);
        }
 }
```

If/when the time comes, let me know which approach you prefer and I'll update the PR accordingly.

avih:
For reference, I'm now in favor of fixing the legacy x11 handler to emit button trigger on button-up.

The top two commits here show this https://github.com/avih/less/commits/mouse-mark-jump-alt
(first commit is unmodified from this PR, second commit replaces the second comit here, and the 3rd commit in this PR is dropped).

I'm keeping the PR unmodified as reference, but I'll update it to this alternative solution if requested.

avih:
Rebased and chose the "fixing the legacy x11 handler to emit button trigger on button-up." approach.

avih:
Rebased after windows key handling refactor.

avih:
Thanks.

Issue #391, opened 2023-06-25
On Windows code points from  Supplementary Planes (U+010000 to U+10FFFF) are not accepted from keyboard


Konstantin-Glukhov:

gwsw:
@adoxa @avih 
I've been experimenting a bit, and the behavior of ReadConsoleInputW seems really strange to me. If I paste a non-BMP char, ReadConsoleInputW return a KeyDown event with AsciiChar and UnicodeChar both set to zero and wVirtualScanCode set to 0x38, then a KeyUp event with UnicodeChar set to the codepoint of the pasted char. 

In contrast, if I press an arrow key, I get a KeyDown event with again AsciiChar and UnicodeChar both set to zero and wVirtualScanCode set to the arrow's scan code, and then a KeyUp event with the fields set the same. This looks very similar to the non-BMP paste.  Currently less processes such an event on the KeyDown, as it should, but to support a non-BMP paste, it seems I would have to wait until I get the KeyUp to distinguish them. This would be a poor user experience.

Is there a way to distinguish a non-BMP char from a regular non-ASCII key like the arrows? The 0x38 scan code seems to be the ALT key. If I press ALT I get a KeyDown event that looks (AFAICT) identical to the first event of a non-BMP char. I don't understand why a non-BMP char doesn't deliver the codepoint in the KeyDown event.

gwsw:
Correction: it's not non-BMP chars that fail to populate the KeyDown event. Some chars work and some don't, and I can't see a pattern. For example, U+00AA works, but all the others between U+00A1 and U+00B4 do not. This is strange.

avih:
It seems there are many issues with `ReadConsoleInputW` (and A), among them missing key-down events, missing key-up events, 0 instead of key-down, and additional issues when pasting unicode and with surrogate pairs.

The issues can also happen with BMP codepoints, see the samples at the comments at the link below.

Additionally, it behaves differently in different consoles (win XP/7/10) and windows terminal (win 10/11).

That's the solution I wrote for busybox-w32, which behaves like `ReadConsoleInputA` which delivers UTF8, while internally it uses `ReadConsoleInputW`:
https://github.com/rmyorston/busybox-w32/blob/d6fb7a2c36f8a6d5fac4754cdfdb5483a03f2892/win32/winansi.c#L1212-L1419

gwsw:
[...]

PR #392, opened 2023-06-25
Fix the issue with Windows code points from Supplementary Planes

Konstantin-Glukhov:
This pull request is to address the issue #391 "Windows code points from Supplementary Planes (U+010000 to U+10FFFF) are not accepted from keyboard"

Please provide feedback.

### DIFF
--- a/screen.c
+++ b/screen.c
@@ -140,20 +140,17 @@ extern int sc_height;
 // UTF16 and UTF8 buffer sizes
 #define UTF8_MAX_LENGTH 4
 #define UTF16_MAX_LENGTH 2
-struct keyRecord
+#define KEY_BUFFER_MAX_LENGTH 6
+struct consoleBuffer
 {
-	int scan;
-	char ascii;
-	char utf8[UTF8_MAX_LENGTH];
-	int utf8_nbytes;
-	int utf8_current_byte;
-} currentKey;
+	char buf[KEY_BUFFER_MAX_LENGTH];
+	int buf_nbytes;
+	int buf_current_byte;
+} console;
+// Scan codes must be returned as 2-byte sequence: with '\0' 1st byte, scancode 2nd byte
+#define setScancode(scancode) (console.buf_nbytes = 2, console.buf[0] = '\0', console.buf[1] = scancode)
 
-static int keyCount = 0;
 static WORD curr_attr;
-static char x11mousebuf[] = "[M???";    /* Mouse report, after ESC */
-static int x11mousePos, x11mouseCount;
-static int win_unget_pending = FALSE;
 static int win_unget_data;
 
 static HANDLE con_out_save = INVALID_HANDLE_VALUE; /* previous console */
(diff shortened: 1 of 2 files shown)


screen.c
```diff
@@ -2853,55 +2854,100 @@ public int win32_kbhit(void)
...
+		if (
+			ip.EventType != KEY_EVENT ||
+			ip.Event.KeyEvent.bKeyDown != TRUE ||
+			(
+			 ip.Event.KeyEvent.uChar.UnicodeChar == 0 &&
+				(
+				ip.Event.KeyEvent.wVirtualScanCode == 0 ||
+				ip.Event.KeyEvent.dwControlKeyState & (RIGHT_ALT_PRESSED|LEFT_CTRL_PRESSED)
+				)
+			) ||
```
gwsw, line 2871: Why was the dwControlKeyState logic changed? I don't really know what the whole purpose of this RIGHT_ALT/LEFT_CTRL thing is.

screen.c
```diff
@@ -2853,55 +2854,100 @@ public int win32_kbhit(void)
...
+			ip.Event.KeyEvent.wVirtualKeyCode == VK_KANJI ||
+			ip.Event.KeyEvent.wVirtualKeyCode == VK_SHIFT ||
+			ip.Event.KeyEvent.wVirtualKeyCode == VK_CONTROL ||
+			ip.Event.KeyEvent.wVirtualKeyCode == VK_MENU
+		) continue;
+
+		// TODO: Document what the following block does
+		if (ip.Event.KeyEvent.dwControlKeyState) {
```
gwsw, line 2879: Why was this conditional added here? That is, why are we doing this stuff only when dwControlKeyState is set?

screen.c
```diff
@@ -2853,55 +2854,100 @@ public int win32_kbhit(void)
...
+		// If we got here, we got a real charachter code
+		if (ip.Event.KeyEvent.uChar.UnicodeChar) {
+			char16_t UnicodeChar = ip.Event.KeyEvent.uChar.UnicodeChar;
+			// If ASCII character, return it without coversion
+			if ( UnicodeChar <= 0x7F ) {
+				currentKey.ascii = ip.Event.KeyEvent.uChar.AsciiChar;
+			// UTF-16 character
+			} else {
+				char16_t utf16[UTF16_MAX_LENGTH];
+				int utf16_nwords;
```
gwsw, line 2923: The utf16 buffer is used when the high surrogate is received and then again when the low surrogate is received, and this code expects it, and utf16_nwords, to be unchanged when the low is received. But these are locals and are reinstantiated between the two calls. I think these two variables should be moved to currentKey, with the rest of the stuff that persists while receiving characters.

screen.c
```diff
@@ -2916,27 +2962,14 @@ public int win32_kbhit(void)
...
-	static unsigned char utf8[UTF8_MAX_LENGTH];
-	static int utf8_size = 0;
-	static int utf8_next_byte = 0;
+	static int pending_scancode = 0;
 
-	if (win_unget_pending)
-	{
-		win_unget_pending = FALSE;
-		return (char) win_unget_data;
-	}
```
gwsw, line 2928: Why was the check for win_unget_pending removed? It looks like WIN32ungetch won't work any more.

Konstantin-Glukhov: This is my screwup. I removed it by accident when I was merging my changes with the latest commit. I put it back.

screen.c
Konstantin-Glukhov: I did not touch the logic, the only thing I did is  moving it into the loop and grouping by dwControlKeyState in one "if" statement to break out of the loop if dwControlKeyState is set.

screen.c
Konstantin-Glukhov: The grouping of all checks is to break the loop (actually return to the caller) if dwControlKeyState  is set. This logic was just moved from the bottom of the function into the loop.

screen.c
Konstantin-Glukhov: utf16 buffer is local because it is used only in one invocation of the function. The reason everything was moved to the loop is to achieve exactly that. The utf16 buffer can be filled only in one call to the function. utf16_nwords can have only two values: 1 (non-surrogate) or 2 (surrogate). This local scope isolate handling non-ascii characters.

screen.c
gwsw: Previously the test was for both RIGHT_ALT_PRESSED and LEFT_CTRL_PRESSED to be set. After your change it only requires either one to be set.

screen.c
gwsw: utf16 is defined in the scope between lines 2921 and 2947. But on line 2926 it sets utf16[0] to the high surrogate and immediately jumps to 2824, which destructs utf16 since 2824 is outside the 2921-2947 scope. Then when the low surrogate is received, it instantiates a new utf16 array at 2922, and stores the low surrogate in utf16[1] on 2930. At that point the code seems to expect utf16[0] to still be populated with the high surrogate. Am I reading something wrong?

gwsw: utf16 could be local, but it has to be defined before line 2824 so that it remains in existence throughout the whole loop.

screen.c
Konstantin-Glukhov: I think I just optimized the code, but please point me where I am wrong. Here is my thinking:
A & (B | C) == (B | C) is equivalent to A & (B | C) == TRUE
To illustrate, let's say A = 1111, B = 0001 and C = 0010
1111 & (0001 | 0010) == 0011
1111 & 0011 == 0011
0011 == 0011
0011 == TRUE

gwsw: Suppose A=0001, B=0001, C=0010.
Then A&(B|C) == (B|C)
is (0001 & 0011) == 0011
is 0001 == 0011 
which is false.

But A&(B|C) == TRUE
is (0001 & 0011) == TRUE
is 0001 == TRUE 
which is true.

screen.c
gwsw: The compiler is not required to destroy an object when it goes out of scope, but it is allowed to do so. See 6.2.4 paragraph 6 [here](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n2310.pdf).

Konstantin-Glukhov: Got it. Thank you for digging this.

screen.c
Konstantin-Glukhov: I will fix it.

Konstantin-Glukhov:
Let me try again from scratch. I will take into consideration all the comments and leave the prior logic untouched except moving everything into the loop.

avih:
Obligatory question: did you test that it works in the windows console (not windows terminal)?

Sample tese case:
- Create a UTF8 text file which contains some text which includes this string: `Ԁ߀ಀ😀` (U+0500 U+07C0 U+0C80 U+1F600)
- Open this file in `less` with this PR patch, press `/` to open the search prompt, paste that string and press enter.

Expected result: that string is found and highlighted at text.


As for the code, note that `keyCount` is not updated when returning UTF8 bytes.

This means that after all the UTF8 bytes are returned, `keyCount` will stay whatever value it was assigned in `win32_kbhit` from `ip.Event.KeyEvent.wRepeatCount`.

This will have several unwanted results:
1. Next time `win32_kbhit` is called, it will NOT read a new console input, because `keyCount` is not 0.
2. Next time `W32getch` is called, it will return whatever was at `currKey.ascii` before the UTF8 was returned - `keyCount` times. This could also be `0`, at which case whoever called `W32getch` will think that the next value is a scan code.
3. `keyCount` is not be respected, for instance if some (non-surrogate pair) unicode value had a count of 2 then `W32getch` will not return the UTF8 bytes twice.
4. Note that with a surrogate pair, `keyCount` of the 1st half is completely ignored. Not sure what the solution is, but I don't know if it should include ignoring the repeat count in some cases.

That's just one thing which caught my eye. There could be other issues too.

avih:
The real problem here is that the keyboard handling is highly delicate code, with multiple states and buffers, each with its own buffer variables and state variables:
- Normal ASCII-7 inputs code wants to return one byte. It's stored at `currKey.ascii`.
- Scan code inputs code wants to return two bytes: 0 and the scan code. It's stored as `currKey.ascii == 0 && curKey.scan == VALUE`.
- The x11 mouse code wants to return 6 values. The first one (ESC) is stored at `currKey.ascii`, and the other five at `x11mousebuf` with their state at `x11mousePos` and `x11mouseCount`.
- The unicode code (which doesn't include ASCII-7...) wants to return 2-4 bytes. The bytes are stored at an array, and there are two state variables (pos and length), which were static at `W32getch` before this PR, and this PR wants to move them to `currKey`.

Additionally, there's `keyCount` which we do want to respect.

Clearly, all of those things want to do a similar thing: enqueue some sequence of bytes, with length of 1-6, and with some repeat count.

But the current implementation is so cumbersome, and touching it in any way is highly likely to introduce state bugs, because the state is so complex.

It wasn't designed that way. Initially it was simple, but over time, it was enhanced further and further, and it reached its current state.

I really do think that before it becomes yet even more complex, all those buffers (including ASCII-7 which can be considered a buffer of 1) should be unified.

And then, once it's possible to look at the code and grasp what it does, then it can be extended further.

Otherwise, these incremental increases in complexity are bound to introduce delicate state bugs, even if they're not identified immediately.

avih:
> I really do think that before it becomes yet even more complex, all those buffers (including ASCII-7 which can be considered a buffer of 1) should be unified.

I'm pretty confident I can do this cleanly. I already had one go at unifying the buffers, but the approach was a bit lacking WRT key count (it was a single bytes buffer and that's it). This can be better (6 bytes buffer and key-count).

I'm willing to spend the time and do that if I get a green light that the maintainers are interested.

avih:
> Please take a look.

Please test that it works first in all the cases which your patch touches.

Konstantin-Glukhov:
@avih, this is not a final commit. As you can notice at the bottom of this page there is an exclamation message "This pull request is still a work in progress". I did not say that the rollback has to stay. I am still working on the code and looking for constructive feedback instead of parental lecturing.

Konstantin-Glukhov:
I think I found the way for the unified buffer to work with holding PgDn key. Please take a look.
I also want to test editchar() function since it depends on win32_kbhit(). Do you know how to trigger its invocation?

gwsw:
This has been fixed as of 010d8086391bdaf291ccd2d523cab683773b2d71.

PR #412, opened 2023-08-13
lesstest: use pkg-config to find ncurses libraries

thesamesam:
Fails to build with split tinfo otherwise like:
```

x86_64-pc-linux-gnu-gcc -O2 -pipe -march=native -fdiagnostics-color=always -frecord-gcc-switches -Wreturn-type      -ggdb3 -Wl,-O1 -Wl,--as-needed -Wl,--defsym=__gentoo_check_ldflags__=0 -Wl,-z,pack-relative-relocs    -ggdb3 -o lesstest display.o env.o lesstest.o parse.o pipeline.o log.o run.o term.o wchar.o -lncurses
/usr/lib/gcc/x86_64-pc-linux-gnu/13/../../../../x86_64-pc-linux-gnu/bin/ld: display.o: in function `display_screen':
/var/tmp/portage/sys-apps/less-643/work/less-643/lesstest/display.c:86:(.text+0x2e0): undefined reference to `tgoto'
/usr/lib/gcc/x86_64-pc-linux-gnu/13/../../../../x86_64-pc-linux-gnu/bin/ld: term.o: in function `setup_mode':
/var/tmp/portage/sys-apps/less-643/work/less-643/lesstest/term.c:83:(.text+0x18): undefined reference to `tgetstr'
[...]
```

Going forward, we may want to just add a Makefile.in which configure handles to the lesstest subdir.

### DIFF
--- a/lesstest/Makefile
+++ b/lesstest/Makefile
@@ -1,7 +1,8 @@
 CC ?= gcc
 CFLAGS ?= -Wall -O2
 LDFLAGS ?=
-TERMLIB = -lncurses
+PKG_CONFIG ?= pkg-config
+TERMLIB ?= $(shell ${PKG_CONFIG} --libs ncurses)
 srcdir ?= .
 
 all: lesstest lt_screen 


uberhacker:
Thank you for this, however, I'm still unable to run tests after applying the patch.  Received the following:
```
ERR  status 256 from /usr/local/tmp/crew/less.20230814224036.dir/less-643/lesstest/lesstest  -e -s '/usr/local/tmp/crew/less.20230814224036.dir/less-643/lesstest/lt_screen' -t '/usr/local/tmp/crew/less.20230814224036.dir/less-643/lesstest/lt/subsearch.lt' '/usr/local/tmp/crew/less.20230814224036.dir/less-643/less'
/usr/local/tmp/crew/less.20230814224036.dir/less-643/less: error while loading shared libraries: libtinfow.so.6: cannot open shared object file: No such file or directory
DIFF github216 on cmd #0 (. 0)
FAIL: github216 (0 steps)
```
The library does exist and the library search path includes this location.
```
$ echo $LD_LIBRARY_PATH
/usr/local/lib:/usr/local/lib64
$ ls -l /usr/local/lib64/libtinfow.so.6
lrwxrwxrwx 1 chronos chronos 16 Jan  5  2023 /usr/local/lib64/libtinfow.so.6 -> libtinfow.so.6.4
```
Any other suggestions?

thesamesam:
I assume `LD_LIBRARY_PATH` is definitely exported (not just set)?

Can you reproduce it if you run `/usr/local/tmp/crew/less.20230814224036.dir/less-643/lesstest/lesstest  -e -s '/usr/local/tmp/crew/less.20230814224036.dir/less-643/lesstest/lt_screen' -t '/usr/local/tmp/crew/less.20230814224036.dir/less-643/lesstest/lt/subsearch.lt' '/usr/local/tmp/crew/less.20230814224036.dir/less-643/less'` manually after failure?

uberhacker:
Hey @thesamesam,
Thanks for the quick response.  Yes, I can reproduce.
```
$ /usr/local/tmp/crew/less.20230817064138.dir/less-643/lesstest/lesstest -e -s '/usr/local/tmp/crew/less.20230817064138.dir/less-643/lesstest/lt_screen' -t '/usr/local/tmp/crew/less.20230817064138.dir/less-643/lesstest/lt/subsearch.lt' '/usr/local/tmp/crew/less.20230817064138.dir/less-643/less'
/usr/local/tmp/crew/less.20230817064138.dir/less-643/less: error while loading shared libraries: libtinfow.so.6: cannot open shared object file: No such file or directory
DIFF subsearch on cmd #0 (. 0)
FAIL: subsearch (0 steps)
```

thesamesam:
I'm not sure what the remaining issue is with their test. I interpreted it as "this doesn't make it worse for me" rather than "this regresses my usecase" though.

But you're right that it won't work if there's no pkg-config program on the build system. Handling that portably in a Makefile is a pain. Any suggestions? The better option would be to have configure look for it and substitute in the value.

avih:
> what will happen if there is no pkg-config program on the build system?

This should handle this case (untested):
```makefile
TERMLIB ?= $(shell ${PKG_CONFIG} --libs ncurses || printf %s -lncurses)
```

gwsw:
This is fixed in a different way in 83bef2bbc3d4f2ca519a9f3973e2513adff44295.

PR #413, opened 2023-08-18
Add padding

miasb:
Added an additional option to pad lines by adding spaces to the line prefix and substracting the requested padding in the fits_on_screen() function.


gwsw:
What is the use case for this option? Why would you want to add blank space at the start of each line?

Also, is the change to fits_on_screen correct? plinestart sets end_column and accounts for the chars added by your new add_pfx calls. fits_on_screen takes end_column into account, so it seems like the subtraction of padding is double-counting the pad chars.

miasb:
For longer texts it is much easier for me to read if its padded. I use it with my use aggregator and html filter to read my news.

I wanted the padding on both sides of the terminal.

Issue #348, opened 2023-03-10
Add at least one example to the lessecho man page


jidanni:
Add at least one example to the lessecho man page.
There may be examples, but they aren't on the man page itself.
```
$ lessecho abc
abc
```
OK, but that is the same as
```
$ echo abc
abc
```
so please add a different example.

gwsw:
lessecho isn't intended to be used by people, only by less itself. I'm not sure it even needs a man page, but I don't think the man page needs instructions for use by humans.

PR #424, opened 2023-09-10
Update lessecho.nro.VER

jidanni:
For #348, so most people know they don't need to learn lessecho.


gwsw:
Fixed in 6efbb89883d339254cd225eb32e850664cc32bc4.

PR #433, opened 2023-09-21
Mingw portability improvements

avih:
Previously, `Makefile.wng` could only be used on Windows, required perl, and with defaults tailored to MSYS2 mingw 64 bits setup - but even that was broken with some weird cmd.exe thing (`make -f Makefile.wng` did not work in MSYS2 mingw 64 env).

Now it supports cross compiling, doesn't require perl (but will use it if it exists), and with a safer default regex package.

Tested on debian 11 (after `suto apt install make mingw-w64 git`), fedora (after `sudo dnf install make mingw32-gcc`), [w64devkit](https://github.com/skeeto/w64devkit) (out of the box), and MSYS2 (with `SHELL=sh`, because the default with cmd.exe is still broken somehow).


avih:
(forced push a trivial indentation change at mkhelp.sh)

avih:
Updated mkhelp.sh again without here-document, which is up to 4x faster in some shells.

avih:
Rebased, updated REGEX_PACKAGE=none to not define NO_REGEX, because `defines.wn` already defines it in such case, so ensure it's not redefined.

Also, updated the commit message of of REGEX_PACKAGE=none and of the change of teh default regex package.

avih:
Fixed negative condition at REGEX_PACKAGE=none, rebased.

avih:
Thanks.

PR #443, opened 2023-10-05
Fixed C99

polluks:
BTW how about including *.uni in .gitignore because it's generated.

### DIFF
--- a/lesstest/run.c
+++ b/lesstest/run.c
@@ -233,7 +233,8 @@ int run_testfile(const char* ltfile, const char* less) {
 }
 
 static void free_states(TestState* states, int num_states) {
-	for (int s = 0; s < num_states; ++s) {
+	int s;
+	for (s = 0; s < num_states; ++s) {
 		free(states[s].screen);
 	}
 	free(states);


avih:
You can also do this, while you're at it:

```diff
diff --git a/command.c b/command.c
index 57cf235..5de9902 100644
--- a/command.c
+++ b/command.c
@@ -704,6 +704,7 @@ static int mca_char(int c)
        case A_B_SEARCH:
                if (incr_search)
                {
+                       int save_updown_match;
                        /* Incremental search: do a search after every input char. */
                        int st = (search_type & (SRCH_FORW|SRCH_BACK|SRCH_NO_MATCH|SRCH_NO_REGEX|SRCH_NO_MOVE|SRCH_WRAP>                        char *pattern = get_cmdbuf();
@@ -714,7 +715,7 @@ static int mca_char(int c)
                         * reinits it. That breaks history scrolling.
                         * {{ This is ugly. mca_search probably shouldn't call set_mlist. }}
                         */
-                       int save_updown_match = updown_match;
+                       save_updown_match = updown_match;
                        cmd_exec();
                        if (*pattern == '\0')
                        {
```

polluks:
@avih This is not required because it's at the beginning of a block.
However another function is missing.

avih:
> This is not required because it's at the beginning of a block.

It's not at the begining. Here's the code which gets patched:
https://github.com/gwsw/less/blob/1ce886241b80a89fa457c0367cf50c3bbbd9e7d4/command.c#L705-L724

avih:
Hre's what I get when compiling with  gcc and `-Wdeclaration-after-statement -std=c89` on Windows:
```
optfunc.c: In function 'toggle_fraction':
optfunc.c:169:17: warning: ISO C90 forbids mixed declarations and code [-Wdeclaration-after-statement]
  169 |                 long tfrac = getfraction(&s, printopt, &err);
      |                 ^~~~
command.c: In function 'mca_char':
command.c:717:25: warning: ISO C90 forbids mixed declarations and code [-Wdeclaration-after-statement]
  717 |                         int save_updown_match = updown_match;
      |                         ^~~
```

gwsw:
Merged and fixed the two warnings raised by @avih.

However, I get many warnings and errors when compiling with -std=c89, because when that flag is specified, <stdio.h> doesn't declare snprintf or popen or PATH_MAX. This is on Fedora 5.12.8-300.fc34.x86_64. Do you see those warnings?

avih:
> Do you see those warnings?

No. When I add to CFLAGS `-Wdeclaration-after-statement -std=c89`, so that each compile line looks like:
```
gcc -c -I. -O2 -DMINGW -DWIN32 -DUSE_REGEXP_C -Wdeclaration-after-statement -std=c89 option.c
```

Both gcc 13.2.0 and clang 17.0.1 (mingw, on windows) only show these two warnings for me.

gwsw:
Hm, I see the extra warnings/errors with -std=c89 on gcc 11.2.1 on Fedora, and on gcc 11.4.0 on cygwin, but not on clang 11.0.0 on Darwin. Well, I guess I could declare those missing functions manually, but I'm not sure I want to go down that route.

avih:
Yeah, even systems which were at their prime in the 80's and 90's many times have at c99 compilers available.

IMO don't touch it, other than the obvious things, like mixed declarations/statements, maybe comments, etc.

c89 is missing many things (especially when it comes to standard types), and it could be a long and deep rabbithole to try actively to make it c89 compliant.

If there's an explicit complaint about some setup, then it can be examined. Otherwise, I don't think it's worth an active effort.

PR #464, opened 2024-01-03
Enable support for IBM z/OS

IgorTodorovskiIBM:
IBM z/OS is a mainframe operating system. 
Changes include:
- z/OS stdout/stderr file descriptors need to be tagged as IBM 1047 to ensure proper conversion
- We pass in an additional library, zoslib, which provides compat with glibc, therefore we need to pass $LIBS in links steps in Makefile.in/Makefile
- Include signal.h for z/OS (`__MVS__`) so that SIGPIPE is available

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -65,10 +65,10 @@ less$(EXEEXT): ${OBJ}
 	${CC} ${LDFLAGS} -o $@ ${OBJ} ${LIBS}
 
 lesskey$(EXEEXT): lesskey.${O} lesskey_parse.${O} xbuf.${O} version.${O}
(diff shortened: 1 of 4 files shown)


IgorTodorovskiIBM:
@gwsw any chance you can look at this PR?

gwsw:
Hi @IgorTodorovskiIBM 
I will look at this soon. I've been working on a large change for the past couple of weeks and don't want to introduce unrelated changes until that one is done. It will probably be another week or so before it's finished.

IgorTodorovskiIBM:
> Hi @IgorTodorovskiIBM I will look at this soon. I've been working on a large change for the past couple of weeks and don't want to introduce unrelated changes until that one is done. It will probably be another week or so before it's finished.

Sounds good, thanks!

PR #466, opened 2024-01-05
handle Linux tracefs files correctly, like procfs

bertschingert:
less already correctly handles files in procfs that claim to have 0 length, but the tracefs filesystem has similar files that report 0 length but have data. This adds support for the Linux tracefs filesystem.

### DIFF
--- a/ch.c
+++ b/ch.c
@@ -718,21 +718,27 @@ public void ch_flush(void)
 
 #if HAVE_PROCFS
 	/*
-	 * This is a kludge to workaround a Linux kernel bug: files in
-	 * /proc have a size of 0 according to fstat() but have readable 
-	 * data.  They are sometimes, but not always, seekable.
-	 * Force them to be non-seekable here.
+	 * This is a kludge to workaround a Linux kernel bug: files in some
+	 * pseudo filesystems like /proc and tracefs have a size of 0
+	 * according to fstat() but have readable data.  They are sometimes,
+	 * but not always, seekable. Force them to be non-seekable here.
 	 */
 	if (ch_fsize == 0)
 	{
 		struct statfs st;
 		if (fstatfs(ch_file, &st) == 0)
 		{
+#if HAVE_TRACEFS
+			if (st.f_type == PROC_SUPER_MAGIC ||
+			    st.f_type == TRACEFS_MAGIC)
+#else
 			if (st.f_type == PROC_SUPER_MAGIC)
+#endif
 			{
 				ch_fsize = NULL_POSITION;
 				ch_flags &= ~CH_CANSEEK;
 			}
+
 		}
 	}
 #endif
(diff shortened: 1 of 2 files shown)


ricardobranco777:
This patch can be improved and simplified by removing the `HAVE_`'s and just not trusting `st_size` at all when it's zero.  It's the only way to also support the Linuxulator on FreeBSD.

bertschingert:
> This patch can be improved and simplified by removing the `HAVE_`'s and just not trusting `st_size` at all when it's zero. It's the only way to also support the Linuxulator on FreeBSD.

Thanks for pointing this out. I definitely agree that it would be better to have a solution that works even when the Linux header isn't available. Another benefit of that is it will remain resilient to future filesystems that may share this behavior.

I uploaded another patch here: https://github.com/gwsw/less/pull/468 which offers another approach to resolving this. The problem with *always* not trusting a zero size from stat() is that the overwhelming majority of the time, stat() doesn't lie about that. Then we do unnecessary read()s and write()s for files that truly are empty.

It occurred to me that less (normally) does a read() right off the bat to check for binary files, and the return value of that read() can be used to check whether stat() is trustable. The advantage of this approach is that it doesn't add any extra syscalls compared to less's current behavior, in the most common case. 

what do you think of this approach?

ricardobranco777:
> > This patch can be improved and simplified by removing the `HAVE_`'s and just not trusting `st_size` at all when it's zero. It's the only way to also support the Linuxulator on FreeBSD.
> 
> Thanks for pointing this out. I definitely agree that it would be better to have a solution that works even when the Linux header isn't available. Another benefit of that is it will remain resilient to future filesystems that may share this behavior.
> 
> I uploaded another patch here: #468 which offers another approach to resolving this. The problem with _always_ not trusting a zero size from stat() is that the overwhelming majority of the time, stat() doesn't lie about that. Then we do unnecessary read()s and write()s for files that truly are empty.
> 
> It occurred to me that less (normally) does a read() right off the bat to check for binary files, and the return value of that read() can be used to check whether stat() is trustable. The advantage of this approach is that it doesn't add any extra syscalls compared to less's current behavior, in the most common case.
> 
> what do you think of this approach?

I created a bunch of PR's for FreeBSD that take the simpler approach of not trusting zero size at all, much like GNU tools do.  If `less` always does a read() then by all means let's take advantage of that.

bertschingert:
> I created a bunch of PR's for FreeBSD that take the simpler approach of not trusting zero size at all, much like GNU tools do. If `less` always does a read() then by all means let's take advantage of that.

Thanks for looking at the PR I posted. Your approach does have the advantage of being a simpler change. Hopefully the less maintainers will chime in on how they want to resolve it.

emaste:
Thanks - I was discussing linprocfs in the FreeBSD context and I wondered how procfs was working in Linux. It seems like avoiding never-ending additions of special cases is going to be the most reliable/future-proof, just never trusting size=0.

gwsw:
Obsoleted by #468.

PR #468, opened 2024-01-09
RFC: use read(), not statfs(), to detect files that lie about size

bertschingert:
Note: Consider this an RFC. I haven't extensively tested this patch for regressions, etc, but am uploading it to see if others agree with the general approach. 

This is another approach to Linux pseudo filesystems, based on this comment: https://github.com/gwsw/less/pull/466#issuecomment-1881135618

-----------

Files in some pseudo filesystems, particularly procfs and tracefs on Linux, lie about their file size in stat(2). They claim to have 0 size, but do return data when read() from.

Previously, less handled these files by using statfs(2) to check the filesystem type (f_type field). However, there are 2 problems with this approach:

- this requires compiling with the Linux "magic.h" header available, and if less is compiled without this header present, it will not have access to the magic numbers that define procfs and tracefs. Such a less binary will not work on these pseudo filesystems
- this approach is not resilient to the addition of new filesystems in the future with similar behavior but a new magic number

This patch takes a different approach; it takes advantage of the fact that less (usually) read()s 256 bytes from the beginning of the file to determine if the file is binary. The return value of the read() can be compared with the filesize from stat(2) to determine if the latter is untrustworthy.

This moves the bin_file() check to happen unconditionally (for regular files, etc.), instead of only happening when the "--force" flag is specified. While this does result in an extra syscall in that particular circumstance, it will result in fewer syscalls overall for the most common scenario of not specifying "--force".

### DIFF
--- a/ch.c
+++ b/ch.c
@@ -20,13 +20,6 @@
 #include <windows.h>
 #endif
 
-#if HAVE_PROCFS
-#include <sys/statfs.h>
-#if HAVE_LINUX_MAGIC_H
-#include <linux/magic.h>
-#endif
-#endif
-
 typedef POSITION BLOCKNUM;
 
 public int ignore_eoi;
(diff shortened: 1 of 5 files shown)


emaste:
> particularly procfs and tracefs on Linux 

mention linprocfs on FreeBSD as well in the commit message, perhaps? a third problem with the statfs check is that it doesn't cover operating systems other than Linux. IMO it makes the case a bit stronger for a change like this rather than the approach in #466

PR #469, opened 2024-01-11
Do not assume PATH_MAX is defined

guillemj:
On systems such as GNU/Hurd, PATH_MAX is not defined, because the system intends to impose no arbitrary limits. In other systems though it might be defined but to a very large value.

We can use realpath() with its POSIX.1-2008 semantics, where passing a NULL argument will make it allocate the destination buffer, but not all systems support these semantics yet.

For now, instead of complicating the code to cope with realpath() limitations on some systems, we simply handle the case where PATH_MAX is not defined, where realpath() should always support these semantics.

### DIFF
--- a/filename.c
+++ b/filename.c
@@ -768,9 +768,24 @@ public char * lrealpath(constant char *path)
 	if (!is_fake_pathname(path))
 	{
 #if HAVE_REALPATH
+		/*
+		 * Not all systems support the POSIX.1-2008 realpath() behavior
+		 * of allocating when passing a NULL argument. And PATH_MAX is
+		 * not required to be defined, or might contain an exceedingly
+		 * big value. We assume that if it is not defined (such as on
+		 * GNU/Hurd), then realpath() accepts NULL.
+		 */
+#ifndef PATH_MAX
+		char *rpath;
+
+		rpath = realpath(path, NULL);
+		if (rpath != NULL)
+			return (rpath);
+#else
 		char rpath[PATH_MAX];
 		if (realpath(path, rpath) != NULL)
 			return (save(rpath));
+#endif
 #endif
 	}
 	return (save(path));


filename.c
```diff
@@ -768,9 +768,24 @@ public char * lrealpath(constant char *path)
 	if (!is_fake_pathname(path))
 	{
 #if HAVE_REALPATH
+	/*
+	 * Not all systems support the POSIX.1-2008 realpath() behavior of
+	 * allocating when passing a NULL argument. And PATH_MAX is not
+	 * required to be defined, or might contain an exceedinly big value.
+	 * We assume that if it is not defined (such as on GNU/Hurd), then
+	 * realpath() accepts NULL.
+	 */
```
jrtc27, line 777:
```suggestion
		/*
		 * Not all systems support the POSIX.1-2008 realpath() behavior
		 * of allocating when passing a NULL argument. And PATH_MAX is
		 * not required to be defined, or might contain an exceedingly
		 * big value. We assume that if it is not defined (such as on
		 * GNU/Hurd), then realpath() accepts NULL.
		 */
```

filename.c
```diff
@@ -768,9 +768,24 @@ public char * lrealpath(constant char *path)
...
+#ifndef PATH_MAX
+		char *rpath;
+
+		rpath = realpath(path, NULL);
+		if (rpath != NULL)
+			return rpath;
```
jrtc27, line 783:
```suggestion
			return (rpath);
```

jrtc27 (commented):
Two style fixes, I believe, from looking at the code (plus a spelling fix)

filename.c
guillemj: Ah! Nice catch, I do recall having that in the my patch in my GNU/Hurd box, but I get that got lost when transplanting it into my main dev box. :)

filename.c
guillemj: Thanks, cice catch also! I was expecting vim spell to be active but for some reason it isn't, will correct that too. :)

PR #471, opened 2024-01-21
Added more formatting attributes for -D

keithbowes:
I was noticing that you can now use --color/-D to set colors, which is so much nicer than the LESS_TERMCAP_<x> environment variables.  However, there was something missing; with the environment variables, I make blinking text bold instead of blinking, but that didn't seem to be possible with -D, so I added it.

The new symbols * and _ for bold and underline are based on common convention/Markdown.  ~ for standout comes from the C not bitwise operator (since terminals tend to reverse the colors in standout mode; I originally made it \ but I thought that might get confused with an escape character; some Markdown dialects use ~ for strikethrough, so ~ might not be the best choice either). & for blink is arbitrary.

This PR however does have one limitation:  It still only works for lowercase letters.  It's still not possible to make the prompt bold, for example.  I might look into why later.

### DIFF
--- a/less.nro.VER
+++ b/less.nro.VER
@@ -679,6 +679,9 @@ the normal format change and the specified color should both be used.
 For example, \-Dug displays underlined text as green without underlining;
 the green color has replaced the usual underline formatting.
 But \-Du+g displays underlined text as both green and in underlined format.
+Other special characters are * for bold mode, _ for underline mode, & for blink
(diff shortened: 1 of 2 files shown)


gwsw:
Thanks. In 608c8e51801a304c7398ba563402a754694f0041 I have implemented this differently, so these new attribute characters are supported for both uppercase and lowercase text types.

PR #476, opened 2024-02-06
Win console title unicode improvements

avih:
The commits are unrelated technically, but related thematically.

The commits reflect the notes and conclusions starting with [this comment](https://github.com/gwsw/less/pull/438#issuecomment-1926946633).

### DIFF
--- a/command.c
+++ b/command.c
@@ -880,7 +880,7 @@ static void prompt(void)
 	{
 		WCHAR w[MAX_PATH+16];
 		p = pr_expand("Less?f - %f.");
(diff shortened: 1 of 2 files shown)


avih:
Force pushed:
- first (no-op) commit: updated the commit message to mention the currently-used non-libc APIs and why it's OK.
- save/restore title commit: added missing space in a comment.

avih:
Hmm.. I see that master was just changed to v650.

FWIW, I think these changes should be included before the next release. They're tiny, safe, and fix issues introduced after the last release.

avih:
Thanks.

PR #478, opened 2024-02-07
reduce less binary size by +2KB by use tabs in less.hlp

kloczek:
Reduce less binary size by use tabs in less.hlp.

$ size less.unexpand less.orig
   text    data     bss     dec     hex filename
 144496   20840   22376  187712   2dd40 less.unexpand
 146896   20840   22376  190112   2e6a0 less.orig

### DIFF
--- a/less.hlp
+++ b/less.hlp
@@ -1,305 +1,305 @@
 
-                   SSUUMMMMAARRYY OOFF LLEESSSS CCOOMMMMAANNDDSS
+		   SSUUMMMMAARRYY OOFF LLEESSSS CCOOMMMMAANNDDSS
 
       Commands marked with * may be preceded by a number, _N.
       Notes in parentheses indicate the behavior if _N is given.
       A key preceded by a caret indicates the Ctrl key; thus ^K is ctrl-K.
 
-  h  H                 Display this help.
-  q  :q  Q  :Q  ZZ     Exit.
+  h  H		       Display this help.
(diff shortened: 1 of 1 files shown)


avih:
One could argue that the help page should be formatted as intended, and unaffected by the terminal tab settings.

Personally I don't think this tradeoff is worth the 1% size saving...

kloczek:
As this will merged I'm going to prepare another PR to use even more tab use to rearrange whole text to use lees spaces ano more tabs so total size reduction will be more than +two times bigger.

gwsw:
The help file deliberately uses spaces rather than tabs, so that text is properly aligned regardless of the user's tab setting. The only way I can see this approach being feasible would be if the code changed the tab setting before viewing the help file and then restored it after exiting the help file. I don't think it's worth this complication to save a few kilobytes.

kloczek:
Your choice ..

avih:
> The only way I can see this approach being feasible would be if the code changed the tab setting before viewing the help file and then restored it after exiting the help file

Well, it's also possible that the content is tabs-to-spaces converted with a tab-equals-8 config, which should be trivial to do for ASCII content (I'm estimating at most 5 lines of code), and fairly simple with existing `wcwidth`-like code otherwise, as I presume `less` already has ways to calculate on-screen width of strings - which is required for tabs-to-spaces conversion.

kloczek:
Q: did anyone tested that using tabs trashes output on DOS/Win? 🤔

PR #488, opened 2024-03-02
Misleading usage syntax

polluks:
The option shift is like line-num-width, match-shift and modelines. Also the brackets are unbalanced.
However the manual says "If L is empty or missing, the number of  header  lines is unchanged." but
```
$ less --header=
Value is required after --header

### DIFF
--- a/less.hlp
+++ b/less.hlp
@@ -212,7 +212,7 @@
                   Set shell quote characters.
   -~  ........  --tilde
                   Don't display tildes after end of file.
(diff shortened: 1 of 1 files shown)


gwsw:
Thanks for pointing out this inconsistency. I've fixed it along with a few others in 7a9753d36f77919939fa555e5ccafdae8c9f0610, so now all the ones whose parameters are optional have brackets around them.

polluks:
fine

PR #500, opened 2024-04-20
Disable "more" compatibility with LESS_IS_MORE=0

pehjota:
A Debian user [reported][1] in 2007 that he has always invoked less as "more" but that the "more" compatibility mode introduced in commit 08a5ba8 didn't work for him.  He requested the ability to disable it, and Debian has carried his patch ever since.  This change is the same patch, forward-ported following commit 5354c92 and with a documentation update.

Example (in source tree):

    $ ln -s less more
    $ LESS_IS_MORE=0 ./more main.c

[1]: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=434417

### DIFF
--- a/less.nro.VER
+++ b/less.nro.VER
@@ -2289,7 +2289,8 @@ In that case, the LESSSECURE and LESSSECURE_ALLOW variables are ignored.
 .
 .SH "COMPATIBILITY WITH MORE"
 If the environment variable LESS_IS_MORE is set to 1,
-or if the program is invoked via a file link named "more",
+or if the program is invoked via a file link named "more"
+and the environment variable LESS_IS_MORE is not set to 0,
 .B less
 behaves (mostly) in conformance with the POSIX 
 .BR more (1)
--- a/main.c
+++ b/main.c
@@ -294,7 +294,8 @@ int main(int argc, constant char *argv[])
 	 * If the name of the executable program is "more",
 	 * act like LESS_IS_MORE is set.
 	 */
-	if (strcmp(last_component(progname), "more") == 0)
+	if (strcmp(last_component(progname), "more") == 0 &&
+			isnullenv(lgetenv("LESS_IS_MORE")))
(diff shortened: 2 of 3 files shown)


gwsw:
I'm not quite following the purpose of this. Why would someone type 
```
$ LESS_IS_MORE=0 more main.c
```
rather than just
```
$ less main.c
```
?

pehjota:
> I'm not quite following the purpose of this. Why would someone type
> 
> ```
> $ LESS_IS_MORE=0 more main.c
> ```

That was a contrived example to show the behavior in one command.  What the user [actually does](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=434417) is set `LESS_IS_MORE=0` in his environment and then for whatever reason "typically invoke the program as 'more'".  I'm not sure why he doesn't just invoke it as "less" (maybe "more" is baked into scripts, "muscle memory", etc.).

But a previous maintainer of Debian's less package accepted the user's patch, so if you don't want to support this (which I'd understand) our options in Debian are to either remove the patch (a regression for anyone relying on `LESS_IS_MORE=0`, so not a good option) or just keep carrying it locally (we'll have to refresh it in the next release due to commit 5354c92, but it's a tiny patch, so not a big deal).  I'm mostly just trying to reduce Debian's local patches to zero (with the other PR merged, this is the last one).

gwsw:
I've implemented this in 8990ba6b967243774a054b154b3619b08e166675 in the post659 branch. I made a small change to make the implementation match the documentation: the documentation says 

>and the environment variable LESS_IS_MORE is not set to 0

but your implementation was instead checking whether LESS_IS_MORE is equal to 1. So in your implementation, setting LESS_IS_MORE=2 would disable more compatibility mode, which seems unexpected. To match the documentation and to avoid unnecessary regression, I changed it so more compatibility is disabled only when LESS_IS_MORE is "0".

PR #501, opened 2024-04-20
doc: Remove claim that "more" can't move backward

pehjota:
As [suggested][1] by a Debian user in 2012.

POSIX [specifies][2] that "more" can scroll backward.

util-linux's "more", added in [version 1.6][3], was copied from [system-0.98.tar.Z][4], where "CREDITS" says it came from BSD Net/2. "more" source code was released in 386BSD version 0.1, upon which NetBSD and FreeBSD were later based.

Both the Net/2-based version in system-0.98.tar.Z and [386BSD 0.1's "more"][5] supported [N]b or [N]^B to skip backward N pages.  In other words, apparently all versions of "more" since its first source code releases allow backward movement.

[1]: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=655926
[2]: https://pubs.opengroup.org/onlinepubs/9699919799.2018edition/utilities/more.html#tag_20_81_13_03
[3]: https://git.kernel.org/pub/scm/utils/util-linux/util-linux.git/tree/Notes.pre1995?id=6dbe3af94#n158
[4]: https://ftp.funet.fi/pub/Linux/tools/system-0.98.tar.Z
[5]: https://github.com/386bsd/386bsd/blob/0.1/usr/src/usr.bin/more/more.c#L1019

### DIFF
--- a/less.nro.VER
+++ b/less.nro.VER
@@ -27,10 +27,8 @@ less \- display the contents of a file in a terminal
 .B Less
 is a program similar to
 .BR more (1),
-but which allows backward movement
-in the file as well as forward movement.
-Also,
-.B less
+but it has many more features.
+.B Less
 does not have to read the entire input file before starting,
 so with large input files it starts up faster than text editors like
 .BR vi (1).


gwsw:
>  all versions of "more" since its first source code releases allow backward movement.

The version of more that I was using in 1983 did not allow backwards movement. That's the reason that I wrote less in the first place; see [the FAQ](http://greenwoodsoftware.com/less/faq.html#history). I think that was on a BSD release but I'm not positive. The BSD code you linked to seems to implement backwards movement by calling fseek, which wouldn't work on a pipe, so perhaps that's the confusion. In any case, I accept that the claim that more does not support backwards movement is now antiquated at best.

pehjota:
**more** was originally written by Eric Shienbrood at UCB in 1980.  It was freely licensed first within Net/1 in June 1989 and wasn't standardized until [Single Unix Specification, Issue 4](https://pubs.opengroup.org/onlinepubs/009656399/toc.pdf) in 1994 (page 4 a.k.a. page 26: "The utilities first introduced in Issue 4 are as follows: [...] more").  Therefore it's unlikely that versions of **more** appeared outside BSD before 1989, so in 1983 you probably used it on BSD.  I did a bit more digging to narrow down some interesting history:

[**more** in 4.2BSD from 1983](https://www.tuhs.org/cgi-bin/utree.pl?file=4.2BSD/usr/src/ucb/more/more.c) (sccsid `"@(#)more.c	4.16 (Berkeley) 83/08/26"`) *lacks* `case 'b':`/`case ctrl('B'):` in `command()`.  This or an earlier version of BSD must be what you used.

[**more** in 4.3BSD from 1986](https://www.tuhs.org/cgi-bin/utree.pl?file=4.3BSD/usr/src/ucb/more/more.c) (sccsid `"@(#)more.c	5.4 (Berkeley) 4/3/86"`) *has* `case 'b':`/`case ctrl('B'):`.  So 4.3BSD was the first release to include a version of **more** that supported backward movement on files (which is how your FAQ describes you using it – on log files, not pipes).

Given that 4.3BSD's **more** and early versions of **less** very coincidentally both use the same `b`/`^B` command keys, I'm guessing that UCB copied the feature from **less**.  I see **less** appeared [in 1994's post-USL-lawsuit 4.4BSD](https://www.tuhs.org/cgi-bin/utree.pl?file=4.4BSD/usr/src/usr.bin/more) (or [perhaps somewhere earlier in 1988?](https://github.com/weiss/original-bsd/commit/273e1e22977e5604989f81b4e288f2e0f744dd37)).

Anyway, on a related note, congratulations on (and thank you for) 40 years of maintaining **less**!

avih:
> ongratulations on (and thank you for) 40 years of maintaining less!

+1024

Many thanks from me as well. This has to be the most used command I use on any system.

gwsw:
It's interesting that in those two early versions of less, the copyright notice has my name spelled incorrectly. So someone other than I must have added those copyright notices. I vaguely remember that when I first gave Berkeley permission to add less to BSD, there was some discussion about licensing (probably because I didn't understand the BSD license) but I don't remember when or how the first copyright notice was added to the files.

Issue #505, opened 2024-05-11
Wrong warning when running on Lear Siegler ADM-5 terminal


fstark:
(Issue reproduced on Linux)

How to reproduce:
Enter ``TERM=adm5 less``

Expected Result:
```
Missing filename ("less --help" for help)
```

Result:
```
WARNING: terminal is not fully functional
Missing filename ("less --help" for help)
```

There is a spurious warning that the terminal is not fully functional.

gwsw:
Fixed in deb498116bd318fc40cee18479413f39e12a6d28.

PR #506, opened 2024-05-11
Fixed upstream #505 : spurious warning on adm5 terminal

fstark:
Fix for issue #505 ("Wrong warning when running on Lear Siegler ADM-5 terminal")

The ``cheaper`` function that compares two ways of using termcap sets ``missing_cap`` to 1 if both ways are unavailable. This function is used to choose between using "al" or "sr" ("add line" or "scroll reverse") to add a line at the top of the screen. None of those are available on an ADM5 terminal.

However, ``less`` already has support for terminal that lacks reverse scroll, and is correctly setting the ``no_back_scroll``, making the warning wrong.

Source code already had a provision to avoid using ``cheaper`` if none of the capabilities were available, but it was only used for OS2. I enabled this for all builds. I don't see any potential for any regression.

### DIFF
--- a/screen.c
+++ b/screen.c
@@ -1431,11 +1431,9 @@ public void get_term(void)
 	t2 = ltgetstr("sr", &sp);
 	if (t2 == NULL)
 		t2 = "";
-#if OS2
 	if (*t1 == '\0' && *t2 == '\0')
 		sc_addline = "";
 	else
-#endif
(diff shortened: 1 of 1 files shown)


gwsw:
Manually merged in deb498116bd318fc40cee18479413f39e12a6d28.

PR #514, opened 2024-05-17
implement env var to offset onescreen cutoff less than full height

smemsh:
The threshold size for `-F`/`--quit-if-onescreen` before paging is activated (and we do prompting) is currently `sc_height - 1`.  This hardcoded size does not allow for those who have a `$PS1` that includes a newline (to separate successive prompts with a blank line), or any other complex multi-line prompt, to use `-F` successfully without losing some lines off the top of the screen at the boundary size.  This should be paged to prevent loss of information (without scrolling back, the need for which would not be obvious since `-F` was used).

`less` does not know the user's prompt height without counting newlines in `$PS1`, which would not work with all shells (such as tcsh).  As a workaround, the submitted patch implements a `$LESS_ONESCREEN_OFFSET` environment variable so the user can inform `less`.  This variable, if a positive integer less than screen size, will add to the offset we look at to see if we're past EOF in `eof_displayed()`, as called by `entire_file_displayed()` while `quit_if_onescreen` is enabled to test if we should `quit()` early.

We have previously calculated the "new" shortened screen size to consider for early quit under `quit_if_onescreen`, by subtracting the provided offset during `get_one_screen()` determination.

### DIFF
--- a/forwback.c
+++ b/forwback.c
@@ -20,6 +20,7 @@ public lbool squished;
 public int no_back_scroll = 0;
 public int forw_prompt;
 public int first_time = 1;
+public int onescreen_offset = 0;
 public lbool no_eof_bell = FALSE;
 
 extern int sigs;
@@ -88,7 +89,7 @@ public lbool eof_displayed(void)
 	 * If the bottom line ends at the file length,
 	 * we must be just at EOF.
 	 */
-	pos = position(BOTTOM_PLUS_ONE);
+	pos = position(onescreen_offset ? BOTTOM_OFFSET : BOTTOM_PLUS_ONE);
 	return (pos == NULL_POSITION || pos == ch_length());
 }
 
@@ -540,10 +541,17 @@ public int get_one_screen(void)
 	int nlines;
 	POSITION pos = ch_zero();
 
-	for (nlines = 0;  nlines < sc_height;  nlines++)
+	int offn, offheight;
+	const char *off = lgetenv("LESS_ONESCREEN_OFFSET");
+
+	offn = off == NULL ? 0 : atoi(off);
+	onescreen_offset = offn < 0 ? 0 : offn < sc_height ? offn : 0;
+	offheight = sc_height - onescreen_offset;
+
+	for (nlines = 0; nlines < offheight; nlines++)
 	{
 		pos = forw_line(pos);
 		if (pos == NULL_POSITION) break;
 	}
-	return (nlines < sc_height);
+	return (nlines < offheight);
 }
--- a/less.nro.VER
+++ b/less.nro.VER
@@ -2503,6 +2503,12 @@ so scrolling may be slower.
 Duration (in milliseconds) after starting to read data from the input,
 after which the "Waiting for data" message will be displayed.
 The default is 4000 (4 seconds).
+.IP LESS_ONESCREEN_OFFSET
+Reduce the number of input lines needed to trigger paging with the \-F
+(\-\-quit-if-one-screen) option.  By default (value 0), paging happens only
+if the input exceeds one less lines than screen height.  Incrementing
+this value decreases that threshold.  Use if your shell prompt contains extra
+newlines, so no unpaged output scrolls off the screen.
 .IP LESS_IS_MORE
 Emulate the
 .BR more (1)
--- a/position.c
+++ b/position.c
@@ -26,6 +26,7 @@ static int table_size = 0;
 
 extern int sc_width, sc_height;
 extern int hshift;
+extern int onescreen_offset;
 
 /*
  * Return the starting file position of a line displayed on the screen.
@@ -45,6 +46,9 @@ public POSITION position(int sindex)
 		break;
 	case BOTTOM_PLUS_ONE:
 		sindex = sc_height - 1;
+		break;
+	case BOTTOM_OFFSET:
+		sindex = sc_height - onescreen_offset - 1;
 		break;
 	case MIDDLE:
 		sindex = (sc_height - 1) / 2;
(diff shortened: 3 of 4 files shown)


smemsh:
For example `PS1='\n \$ '` with `LESS_ONESCREEN_OFFSET=1` would give correct behavior.

smemsh:
@gwsw any chance of getting this merged? or objections that I can try to address?

gwsw:
Right now I'm leaving the master branch untouched except for simple bug fixes. I plan to do a bug fix release from master in a few weeks. New development is happening in the post659 branch. I haven't dealt with this PR yet because I've never merged a PR into a branch before and I haven't yet looked into how to do that. I also want to analyze the impact of this change carefully in light of your last comment on May 17.

smemsh:
Thanks for your reply, I do concur that someone more familiar with the full program should verify implications for other users of `eof_displayed()`, given the large installation base of `less`.  I gave it a fair amount of thought, but am not fully certain.

The patch has been rebased off `post659`.  I will monitor upstream commits and keep it based off the branch tip, so it can be applied easily as a fast-forward when you get around to it.  Should you need it moved to a different branch later, depending on timeline, let me know.

Please take your time; I am happy to know the patch is being considered.

gwsw:
I've implemented this slightly differently in 1d36fbd0ab607ea7f09ceef8fa6a2da618876b29. First, I changed the environment variable from LESS_ONESCREEN_OFFSET to LESS_SHELL_LINES, whose value is one higher. I think it's easier for a user to remember that the environment variable should be set to the number of lines in their shell prompt rather than one less than that. I also changed eof_displayed() to use this value only when called from entire_file_displayed(). I ran into at least one issue caused by using the offset value indiscriminately: if you run ```seq $LINES | less``` and page down, the prompt doesn't change to ```(END)``` at the end of the file.

smemsh:
Works for me, thanks.

vincentbernat:
It works for me too, but not when used with `-X`.

smemsh:
@vincentbernat it works for me with `-X`, for example, using:
```
$ export PS1=$'\n\x20\$\x20' LESS_SHELL_LINES=2
```
the following does not page:
```
for ((i = 1; i < LINES - 1; i++)); do echo $i; done | LESS= less -XF
```
but when LINES is used (without `-1`) it does page.  This seems right to me, do you have a reproducer?

gwsw:
In that example it should page, but I can reproduce that it behaves differently with and without the -X flag.

This pages:
```
seq $((LINES-1)) | LESS_SHELL_LINES=2 less -F
```

This does not page:
```
seq $((LINES-1)) | LESS_SHELL_LINES=2 less -XF
```

I am investigating.

gwsw:
Fixed in 56fb53f2e15ad5fe58577b9fc7b99de0e3b33318.

smemsh:
I'm confused because in my example, either with or without `-X`, the behavior is correct using your old version.  Namely, that the transition point has line 1 at the top of the screen after subsequent prompt regenerated and no paging had activated, but adding one more to the loop activates paging.  Our examples differ by one because of my `<` rather than `<=`, but using your `seq` example, the last argument value that's unpaged should be `LINES - LESS_SHELL_LINES` right? Resulting in a completely filled screen to the top line, after the new prompt is written.  Adding one more (`seq $((LINES - (LESS_SHELL_LINES - 1)))` and paging triggers.

It works that way in your old version both with and without `-X`.  Your new version works the same.  I see no difference in behavior, all 4 variations (with/without -X, 994786e (old) vs 56fb53f (new)) are working correctly (and behave identically). :thinking:

smemsh:
btw just to summarize the difference I have with you, your last example does not match what I see:

> This does not page:
> ```
> seq $((LINES-1)) | LESS_SHELL_LINES=2 less -XF
> ```

For me, that *does* page with 994786e, using a PS1 with a single newline inside.  Change to `LINES - 2` and it doesn't page anymore.  Same behavior seen without `-X`.

gwsw:
Well I don't know how to explain that. I'm not sure it matters much, but if you want to debug it, you could run 994786e4148efbadc89a15138b1dfcf3881166ea  in a debugger and set a breakpoint in get_one_screen(). That should not get called when -X is used, and if it's not called then the LESS_SHELL_LINES variable is not used.

smemsh:
Yeah so it's not breaking, never enters `get_one_screen()`.  Nonetheless paging is activated with a file made with `seq $((LINES - 1))` (have to use `run < testfile` in gdb) with a PS1 having one extra newline in it.

I'm not sure if I trust the test under gdb though because even with `-X` it doesn't leave the screen contents on the screen after gdb exits, even when I use batch mode.  I can see the right options in `show args` and correct environment using `getenv()` shown at a `main()` break.  I think it does its own tty stuff and I'd have to run gdb using a different control terminal than the inferior process.  The manual claims gdb restores the terminal in between prompts, but something is different...

Note that, using `-FX` (no gdb), `less` has the correct cutoff even with my system's `/usr/bin/less`, which is v590.  So my guess is it was off by one before there were any changes, but only if `-X`, for some reason.  This happens to work correctly if PS1 also contains one newline (for a total of 2 newlines generated at the prompt).  It's not reading LESS_SHELL_LINES in that version, of course.

Anyways, I think you're right that it's not worth further investigation, since the 659 trunk behaves correctly in all cases.  Thanks for going through it.

smemsh:
@gwsw noticed `$LESS_SHELL_LINES` was not in `v668-rel`.  I had thought `post659` was getting merged for the next stable release after `v661-rel`, which at the time was pending and we didn't want to put new features it, but would wait for the next one (`v668-rel`).  did I misunderstand the release strategy?

gwsw:
v668 is a bug-fix release on v661. v668 has no new features beyond what is in v661. Now that v668 is released, I will merge post659 into master, probably in about a week or two. So the next production release after v668 will include what is in post659.

PR #524, opened 2024-06-02
fixes #235

### LINKED ISSUE #235: Infinite loop in the help
Run `less --lesskey-src=./foo ./foo` where
```
$ cat foo
b set-mark bg
```
Pressing "b" works correctly but pressing "h" and then "b" results in an infinite loop.  It is okay for `set-mark` to be ignored in help mode, but then its following letter should also be ignored.

cero1988:
fixes #235
dont unget multiple commands in help menu

### DIFF
--- a/less.nro.VER
+++ b/less.nro.VER
@@ -1332,10 +1332,10 @@ will display the contents of that new file.
 .RS
 Sets the number of header lines and columns displayed on the screen.
 The number of header lines is set to \fIL\fP.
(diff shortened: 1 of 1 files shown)


gwsw:
While this does fix #235, I'm not sure it's reasonable. It may break features that the user has created in their lesskey file, and expects to work while viewing the help screen, like a shortcut key to search for a particular string, or to toggle a particular option. It would be unexpected and certainly a regression for such shortcuts to stop working in the help screen.

PR #527, opened 2024-06-10
windows: don't search config files in CWD and $PATH

avih:
Config files, like _lesskey, are determined at homefile(...), and on windows, after trying $HOME and $HOMEDRIVE$HOMEPATH, it could fallback to using _searchenv, which first searches CWD and then in $PATH.

Disable this fallback on Windows.

Note that OS2 and DOS variants (DJGPP, Borland, MS-DOS-compiler) still behave like before, and search PWD and $PATH (assuming they still compile - all the non-DJGPP targets use and used _searchenv, which probably only exists on WIndows).

At the very least the DOS variants don't have a notion of $HOME, so it's not obvious where the config files should be searched - if we wanted to address these build targets as well, but for now we don't.

### DIFF
--- a/filename.c
+++ b/filename.c
@@ -288,7 +288,7 @@ public char * homefile(constant char *filename)
 	if (pathname != NULL)
 		return (pathname);
 #endif
-#if MSDOS_COMPILER || OS2
+#if (MSDOS_COMPILER && MSDOS_COMPILER!=WIN32C) || OS2
 	/* Look for the file anywhere on search path. */
 	pathname = (char *) ecalloc(_MAX_PATH, sizeof(char));
 #if MSDOS_COMPILER==DJGPPC


avih:
Updated the commit message without changes to the code.

gwsw:
Manually merged into the post659 branch in 3bcd6c2262134e92eca473fcadb0124f9586ae26.

avih:
> Manually merged

Did you see my comment two weeks ago here? https://github.com/gwsw/less/commit/1ca819b337894f349b685c1fce18bd4cca5df33d#commitcomment-144844485 when you accidentally merged it into master?

Because it again has the same issue that the commit title dissappeared and each paragraph at the commit message became one line.

gwsw:
Yes I saw that. I did the merge into post659 in 3bcd6c2262134e92eca473fcadb0124f9586ae26 on July 24 which was before your comment. Are you suggesting that I amend the commit message?

avih:
> I did the merge into post659 in [3bcd6c2](https://github.com/gwsw/less/commit/3bcd6c2262134e92eca473fcadb0124f9586ae26) on July 24

Oh, I thought you merged it just before you posted the message.

No, don't ammend it. Sorry.

PR #530, opened 2024-06-13
fix #235

### LINKED ISSUE #235: Infinite loop in the help

cero1988:
fix #235

### DIFF
--- a/command.c
+++ b/command.c
@@ -2085,6 +2085,7 @@ public void commands(void)
 			 * Set a mark.
 			 */
 			if (ch_getflags() & CH_HELPFILE)
+				c = getcc();
 				break;
 			start_mca(A_SETMARK, "set mark: ", NULL, 0);
 			c = getcc();
--- a/command.c
+++ b/command.c
@@ -2085,8 +2085,10 @@ public void commands(void)
 			 * Set a mark.
 			 */
 			if (ch_getflags() & CH_HELPFILE)
-				c = getcc();
+			{
+				getcc();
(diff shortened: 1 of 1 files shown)


gwsw:
The problem I see with this commit is the user experience is poor if the user tries to manually set a mark in the help file. Any command entered after a mark command is ignored:

1. Run less.
2. Press ```h```. The help page is displayed.
3. Press ```m```. The mark command is ignored as it should be.
4. Press any letter representing a valid command, such as ```q``` (to exit help). It is silently ignored, which is likely to be unexpected and confusing.

gwsw:
This modification of your patch might work better, but I haven't fully tested or analyzed it yet.
```
@@ -2085,7 +2085,11 @@ public void commands(void)
             * Set a mark.
             */
            if (ch_getflags() & CH_HELPFILE)
+           {
+               if (ungot != NULL)
+                   getcc();
                break;
+           }
            start_mca(A_SETMARK, "set mark: ", NULL, 0);
            c = getcc();
            if (is_erase_char(c) || is_newline_char(c))
```

gwsw:
Fixed in f0eb3ee233ef88b224895e292b053441b8a067b1.

PR #539, opened 2024-07-02
windows 10+: make 256/true colors "work" without -Da

### LINKED ISSUE #538: colors in bat default theme stopped working in v661 on Windows, but worked in v643
See https://github.com/sharkdp/bat/issues/1573

Colors printed by [`bat`](https://github.com/sharkdp/bat) using just `-R` used to work in less v643, but no longer do in v661 without `-Da` which was suggested as a workaround.

image

image

avih:
See the commit message for full description.

TL;DR:

Before commit f9f2498f (which is just before v651), unknown SGR values resulted in aborting the SGR parsing and instead passthrough till the rest of the buffer to the terminal directly.

This had broken output if the terminal doesn't support VT (e.g. win7), but seemingly correct output with VT, e.g. 256 colors on win10.

Commit f9f2498f changed this process, by ignoring only the rest of the current sequence if an unknown value is encountered, so that the next sequences in the buffer are still processed normally.

Additionally, we then never lose track of the terminal state, because either we recognize a value and apply it, or we do neither of those. Specifically, we never apply a value which we can't recognize/track.

While it fixed the junk output when VT is disabled, it also had an effect which made 256 and true colors stop "working" on win10+.

So with the latest release v661, it can be observed than 256 and true colors stopped working compared to the previous release v643.

So this commit brings back this behavior of "working" 256/true colors on win10+ by default, but a bit more refined than before f9f2498f .

Fixes #538

### DIFF
--- a/output.c
+++ b/output.c
@@ -254,6 +254,21 @@ static void win_flush(void)
 		char *anchor, *p, *p_next;
 		static t_sgr sgr;
 
+		/* when unsupported SGR value is encountered, like 38/48 for
+		 * 256/true colors, then we abort processing this sequence,
+		 * because it may expect followup values, but we don't know
+		 * how many, so we've lost sync of this sequence parsing.
+		 * Without VT enabled it's OK because we can't do much anyway,
+		 * but with VT enabled we choose to passthrough this sequence
+		 * to the terminal - which can handle it better than us.
+		 * however, this means that our "sgr" var is no longer in sync
+		 * with the actual terminal state, which can lead to broken
+		 * colors with future sequences which we _can_ fully parse.
+		 * in such case, once it happens, we keep passthrough sequences
+		 * until we know we're in sync again - on a valid reset.
+		 */
+		static int sgr_bad_sync;
+
 		for (anchor = p_next = obuf;
 			 (p_next = memchr(p_next, ESC, ob - p_next)) != NULL; )
 		{
@@ -289,6 +304,7 @@ static void win_flush(void)
 					anchor = p_next = p;
 					update_sgr(&sgr, 0);
 					set_win_colors(&sgr);
+					sgr_bad_sync = 0;
 					continue;
 				}
 				p_next = p;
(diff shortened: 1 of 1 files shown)


avih:
Rewrote the commit message to explain how v661 can be considered a regression compared to v643 in some cases, and how this commit does the best of both worlds to a degree which it appears perfect.

avih:
So, I started doogfooding master just now, and noticed that there's an issue, possibly related to this commit.

Less crashes with this file (16 lines true color "plane") before completing two lines - [true-color.zip](https://github.com/user-attachments/files/16325540/true-color.zip) .

Each line is ~ 1400 bytes of true color escape sequences, and ends in SGR reset.

I suspect a buffer overflow issue which got exposed by this PR, but I don't have more details currently.

I'll debug and post updates here.

Meanwhile, please wait withthe release.

avih:
The offending commit is mine - f9f2498f20fa2f8f6126006eb19e7e6fc157739b just before v651.

Debugging.

avih:
OK, I still don't know why it crashes, but this is not a new issue.

It was exposed by this part of the diff of commit f9f2498f20fa2f8f6126006eb19e7e6fc157739b :
```diff
@@ -299,16 +312,20 @@ static void win_flush(void)
                                        }

                                        if (q == p ||
-                                               code > 49 || code < 0 ||
                                                (!is_ansi_end(*q) && *q != ';'))
                                        {
+                                               /*
+                                                * can't parse. passthrough
+                                                * till the end of the buffer
+                                                */
                                                p_next = q;
                                                break;
                                        }
```

I.e. this code aborts the parsing and switches to passthrough based on the inaccurate test of `code > 49 || code < 0`.

And that commit ignores the rest of the sequence instead of switching to passthrough.

I tested also with commit 52b8e35 - which is 1 commit before we started touching `win_flush`. By default it doesn't crash, but if I comment out that `code > 49 || code < 0 ||` line then it crashes.

I still don't know why, but I suspect it's an edge case bug which my new SGR parser fixes (it doesn't crash with my own from-scratch parser, and this is why I haven't noticed it, because in my build I use my parser, but I wanted to dogfood master a bit before your release, and I discovered it).

@gwsw any thoughts?

I still have a hunch about buffer overflow, but that test line (>49, <0) is just a trigger. There's nothing special about these values, and these thresholds are arbitrary, and there are unrecognized numbers even within this range (like 38), and the code implicitly does SGR reset in such case - it assumes any unrecognized value must have been 0:
```c
    default:
    /* case 0: all attrs off */
```

So there's nothing special about values outside this range which would trigger bad behavior.

Amd I think it has to be the fact that it doesn't abort early, combined with this specific input.

But I don't yet know why that results in a crash.

avih:
It's "fixed" If I change:
```c
static char obuf[OUTBUF_SIZE];
```
into
```c
static char obuf[OUTBUF_SIZE * 2];
```
I.e. from 1024 to 2048 if I got the numbers right.

avih:
OK, I sort of got it.

This loop:
```c
		for (anchor = p_next = obuf;
			 (p_next = memchr(p_next, ESC, ob - p_next)) != NULL; )
```
can be entered (I'm guessing not in the first iteration) when `p_next` is past the end of the buffer (`ob`), and then all bets are off.

Now only remains to find the offensive asumption or calculation inside that loop.

This shouldn't take long.

PR #546, opened 2024-07-21
windows: advance more carefully while parsing SGR

avih:
The SGR processor parses an input buffer for SGR sequences, and not
all the places which advanced the data pointer checked whether it
reached the end of the buffer (EOB - first byte past the buffer).

If it did reach the EOB and we didn't realize it, then it could
result, depending on how the stars align, in a crash (because the
code updates global pointers with some prtdiffs, and if the diff
is negative because we weren't careful then all bets are off).

With a specific test file which resulted in a crash, the first
offending advancement which reached EOB untested was:
```c
  if (*q == ';')
      q++;
```
However, looking at the code, there seem to be more places where
potentially bad advancements were not protected.

This commit adds several advancement protections, and fixes the crash.

It's possible that it's over-careful and that some of the tests are
always "no problem", or even that all of them are not required except
that one "q++" advancement which we know can reach EOB.

But this code is complex, and it's better to be on the safe side.
If a test is always "no problem", then no problem, it's no-op.

This addresses the issue mentioned at https://github.com/gwsw/less/pull/539#issuecomment-2241727530

### DIFF
--- a/output.c
+++ b/output.c
@@ -236,6 +236,12 @@ static void set_win_colors(t_sgr *sgr)
 	WIN32setcolors(fg, bg);
 }
 
+/* like is_ansi_end, but doesn't assume c != 0  (returns 0 for c == 0) */
+static int is_ansi_end_0(char c)
+{
+	return c && is_ansi_end((unsigned char)c);
+}
+
 static void win_flush(void)
 {
 	if (ctldisp != OPT_ONPLUS
@@ -294,7 +300,7 @@ static void win_flush(void)
 					anchor = p;
 				}
 				p += 2;  /* Skip the "ESC-[" */
-				if (is_ansi_end(*p))
+				if (is_ansi_end_0(*p))
 				{
 					/*
 					 * Handle null escape sequence
@@ -313,7 +319,7 @@ static void win_flush(void)
 				 * Parse and apply SGR values to the SGR state
 				 * based on the escape sequence. 
 				 */
-				while (!is_ansi_end(*p))
+				while (!is_ansi_end_0(*p))
 				{
 					char *q;
 					long code = strtol(p, &q, 10);
@@ -333,7 +339,7 @@ static void win_flush(void)
 					}
 
 					if (q == p ||
-						(!is_ansi_end(*q) && *q != ';'))
+						(!is_ansi_end_0(*q) && *q != ';'))
 					{
 						/*
 						 * can't parse. passthrough
@@ -355,7 +361,7 @@ static void win_flush(void)
 
 					p = q;
 				}
-				if (!is_ansi_end(*p) || p == p_next)
+				if (!is_ansi_end_0(*p) || p == p_next)
 					break;
 
 				if (sgr_bad_sync && vt_enabled) {
--- a/output.c
+++ b/output.c
@@ -79,7 +79,17 @@ public void put_line(void)
 	at_exit();
 }
 
+#if MSDOS_COMPILER==WIN32C || MSDOS_COMPILER==BORLANDC || MSDOS_COMPILER==DJGPPC
+/*
+ * win_flush has at least one non-critical issue when an escape sequence
+ * begins at the last char of the buffer, and possibly more issues.
+ * as a temporary measure to reduce likelyhood of encountering end-of-buffer
+ * issues till the SGR parser is replaced, use bigger (8K) buffer.
+ */
+static char obuf[OUTBUF_SIZE * 8];
+#else
 static char obuf[OUTBUF_SIZE];
+#endif
 static char *ob = obuf;
 static int outfd = 2; /* stderr */
 
--- a/output.c
+++ b/output.c
@@ -332,8 +332,7 @@ static void win_flush(void)
 						 * in the buffer.
 						 */
 						size_t slop = ptr_diff(q, anchor);
-						/* {{ strcpy args overlap! }} */
-						strcpy(obuf, anchor);
+						memmove(obuf, anchor, slop);
 						ob = &obuf[slop];
 						return;
 					}


avih:
Updated the commit message, no code changes.

It turns out that two of the protections together are required to fix the crash. If only one of them has protection, it still crashes. (the code adds 4 protections, two of which now seem mandatory).

avih:
Well, I was trying to reduce the protections by following the code closely, and I found out something disturbing:

`is_ansi_end(0) == 1`.
(because `is_ascii_char(0) == 1`, and `strchr(end_ansi_chars, 0) != NULL`)

This was one of the reasons for going past the end of the buffer, and possibly the only reason.

One of the increments reached the final `\0`, and it incorrectly thought it's a valid end of SGR because `is_ansi_end(0) == 1` but it's already past the end of the input, and it incremented it again for the next iteration, and this was already past the final `\0' and from there things can only go down.

I feel like logically `is_ansi_end(0)` really should not say "yes", but I think this is not something we want to change in a bugfix release, because who knows what depends on it.

And indeed, if I change every `is_ansi_end(x)` in `win_flush` into `(x && is_ansi_end(x))` (there are 4), and remove all the protections, then it doesn't crash.

In this different "fix" I didn't yet follow the code closely to detect whether some protections are still required.

But I think whoever wrote the code definitely didn't take into account that `is_ansi_end(0)` is "yes" ...

Not sure how to proceed.

Do you want to take this one yourself? I didn't write this code and I'm not enjoying trying to follow it, it's really hard to follow the variables (even after we moved the actual SGR values processing into a different function).

I also confirmed that this fix (`is_ansi_end(x)` --> `(x && is_ansi_end(x)` in the 4 cases it's tested) also fixes 52b8e35df9cb1d55359f169e4f4379edc539639f (after commenting out the line which hidden this bug: `code > 49 || code < 0 ||`).

I.e. before we started touching `win_flush`, it crashes after commenting out `code > 49 || code < 0 ||`, and applying these 4 `is_ansi_end` changes fixes it. So it's definitely note a regression of the refactor.

`is_ansi_end` was added in 2001 in commit f3e54d033c476cd576727e49094b00ac0a30b399 and already back then it returned 1 for input 0, so it's not something which changed in recent decades and broke the code.

I think the author simply assumed that it would return 0 for input 0, and built the logic around that assumption and that's it.

I use less quote a lot, but for years I was using an old msys1 versions, which I don't know if used that `win_flush` code.

But once I moved to my own build, one of the first things I did was writing my own SGR parser, and I've been using that ever since, so I never noticed this issue.

And the first time I started testing master today, it came up pretty quickly.

I'm not sure how it wasn't noticed till now, and I think the bug should really still be there even if we don't remove the `code > 49 || code < 0 ||`, but I don't know how to reproduce it in such case.

gwsw:
I agree that it is unexpected and disturbing that is_ansi_end(0) returns true. I also agree that changing it in a bug fix release is probably not desirable, but I'm not sure about that and I will think about this some more. For now, would it be reasonable to change the 4 calls to is_ansi_end in win_flush to just compare to 'm'? Setting LESSANSIENDCHARS must be exceedingly rare, especially on Windows, so I can't imagine that breaking anything.

avih:
> would it be reasonable to change the 4 calls to is_ansi_end in win_flush to just compare to 'm'?

This should have effectively the same effect of changing `is_ansi_end(x)` into `(x && is_ansi_end(x))` but less configurable (but I agree no one is likely to want anything other than `m`), and I don't think they pose different risk levels, so between those two I'd choose my own suggestion.

But yours is fine as well, and I don't feel strongly about it.

However:
> In this different "fix" I didn't yet follow the code closely to detect whether some protections are still required.

It's possible that it only happens to fix the crash in this specific test case, and that some protections are still required, but it's not easy to follow the code in order to know that for a fact.

OTOH, the protections are also a bit of hand waving, becaue I didn't initially realize how come it goes past the end of buffer, and it was enough to protect each increament - but still a bit dodgy, and I don't feel 100% comfortable with that either.

The only "good" way to address it is to follow the code closely and conclude what is required exactly, if indeed anything is still required beyond fixing `is_ansi_end`.

avih:
OK. I think I understand why removing the line `code > 49 || code < 0 ||` triggers the issue, and why changing the buffer size from 1K to 2K "fixes" it.

In the vast vast majority of cases "less" doesn't have to use more than a single buffer (1K) to contain the full output for one line, so all the code which deals with reaching the end of the buffer (and its consequences of moving the remainder to the begining) was never tested in practice

But this test file is different. It has 64 columns (of spaces) in each line, but each space is in a different true color background color in the form of `ESC [ 48 ; 2 ; 123 ; 234 ; 255 m`, which happens to be about 1400 bytes per line.

However, with that trigger line still in place, it still doesn't have to use the EOB code, becuse it switches to passthrough till EOB when it encounters the `123` in the example above (123 > 49), so there's no need for the delicate EOB code to get praciced.

But with that line removed, it means it doesn't switch to passthrough anymore (at least for unknown SGR values), and the parsing continues till EOB, and then fails to handle EOB correctly., simply because it was not debugged enough because it never happened.

Now that we added 256/true color "support" and it actually works (v662), and it doesn't passthrough the rest of the buffer when it encounters e.g. 123 (v644, right after v643 was released), then it means the "end of buffer reached" code can be entered on a more regular basis (but probably still not frequently, because not many files change true color every column, but it can still definitely happen).

I can think of few things.
1. Indeed make sure that is_ansi_test or equivalent doesn't consider 0 as ansi-end.
2. Increase the buffer. It's 1k, currently, so, dunno, make it 8k, or 32k, or whatever. It's static, no problem with that, and it would make it much more likely to never reach the end of buffer in a single line and use code which didn't get much testing (and we can restore it once we switch to my parser - which I did test with tiny buffer, and which always consumes the whole buffer without this "move the remainder to the begining for future re-processing" behavior).
3. I still want to follow that code closely, and hopefully I'll have something concrete to say about it.

gwsw:
>Change that strcpy to move the remainder into memmove.

If this would fix the buffer overrun, it seems like the safest change for the bug fix release. I will fix is_ansi_end(0) in the post659 branch but I think it would be best not to introduce that into the bug fix release.

avih:
OK, it's painful, but I'm getting the hang of it.

Few data points:
- The original code at commit c2709ff87b23b1da99ae46b785543fa085bbcbda either definitely assumed that `is_ansi_end(0)` is 0, or it was very careless. On entry to this code `*p` is ESC:
```c
    if (p[1] == '[' && is_ansi_end(p[2]))
    {
        textattr(norm_attr);
        p += 3;
        anchor = p_next = p;
        continue;
    }
```
And when `p[2]` is `\0` at EOB and this code is entered because is_ansi_end, then p+3 is past the `\0` and it's downhill from there. So it likely assumed p[2] == 0 is not ansi-end.
- I fixed is_ansi_end in this function (and only this function) to return 0 if the value is 0, and changed the buffer size to 8 (7 for text, one for `\0`), and tested quite a lot with syntax highlighted content which has at most 7-bytes sequences (e.g. ` ESC [ 33 ; 1 m`) and it didn't crash, which is great. Surprisingly I also didn't notice performance difference.
- I added debug output to a file when the "incomplete sequence, move the remainder to the begining" thing happens, and indeed it happens all the time (with buffer size 8), and the output looks fine, and there were (tenths of) thousands of those. So I think the EOB code is quite likely not crashy when is_ansi_end is fixed.
- I logged also the size of the reminder which is moved on incomplete sequence, and noticed all the values between 2-6 happening many times, but never 1. Looking at the code, there's indeed a bug that if the last char at the buffer is ESC, then it's not considered incomplete sequence. Instead it's accidentally considered normal text (or rather ESC-not-[), and copied to the output as text instead of getting processed as SGR at the next call. On win10 that's fine because the terminal ends up processing this sequence, but on win7 there's no SGR processing of this, and buggy output. This is easy to fix.
- The variable `p_next` definitely changes semantics throught the code, as if the author wanted to save variable names and reuse existing variables for different things. Also at least two times it's assigned a value which can never be used (right before the end of the loop, where if the loop is entered then it's assigned unconditionally, ignoring the last assignment, and if the loop is not entered again then it's not used outside the loop either). This is the most annoying variable at this code.

That's it for now. Still trying to analyze this unholy function.

avih:
> If this would fix the buffer overrun

I didn't notice an issue with `strcpy` still at the code at the experiments I reported at my comment above (but it should still be memmove)

gwsw:
Ok, sounds like you're making good progress. Let me know when you have a solution to all these issues. I won't merge the PR until you're satisfied that everything has been resolved.

avih:
> I will fix is_ansi_end(0) in the post659 branch but I think it would be best not to introduce that into the bug fix release.

I agree that in general it shouldn't be fixed for this bugfix release, but the code in win_flush has to do something about it regardless.

I was thinking adding this to output.c:
```c
/* like is_ansi_end, but doesm't assume c != 0  (returns 0 for c == 0) */
static int is_ansi_end_0(char c)
{
	return c && is_ansi_end((unsigned char)c);
}
```

And use that from win_flush. That's OK, yes?

gwsw:
Yes, that seems fine to me.

avih:
> p_next is the most annoying variable at this code.

Indeed. It's used all over the place and looks like the most important variable, but it's nigh-impossible to grasp what its role is.

That's because it has two completely unrelated roles, which are additionally unnecessary.
<details>
<summary> The roles of p_next: </summary>

- Hold the origin and result of finding the next ESC char at the begining of each "for" iteration. This role ends at the very begining when the result is assigned to `p`, then all the work is done with `p`, and then this role is resumed at the end of the iteration when p_next is assigned from the updated `p` for the next iteration. Between these two edges p_next is not used in this role and its value doesn't matter. This role can be eliminated by simply using `p` to find the next ESC instead of this juggling.
- Because the value of p_next inside the "for" loop doesn't matter, it's abused to help with a completely unrelated task of implementing "break 2" (bailing out of two nested loops) when SGR syntax error is detected. This role begins with the init `p_next = p` at the middle of the "for" loop, and ends at the `break` out of the "for" loop. Once the outer ("for") loop is bailed, p_next is not used further, the remainder of the buffer is flushed, and the function exits. This role can be eliminated by simply flushing the rest of the buffer and exiting as soon as the syntax error is detected.

With that knowledge, it's trivial to eliminate p_next at the code and make it more readable, but currently p_next is almost guaranteed to be a red herring in any code analysis attempt.
</details>

However, back on topic, these p_next roles are harmless, and have no effect on the end-of-buffer behavior.

Moving on.

There are 3 places where a pointer inside the buffer is incremented explicitly (`p += 2`, `p++`, and `q++`), and in all cases it can't get past the final `\0` as far as I can tell - but only if `is_ansi_end(0) == 0`  (it can reach the `\0`, but the code eventually detects it and saves it for later without reading past the `\0`).

There's another advancement using `strtol` which is also guarded correctly as far as I can tell.

And the main advancement using `strchr(s, ESC)` which is also OK.

These, together with my experiments with a tiny buffer and many colors, support the assessment that once `is_ansi_end` is fixed, the code becomes safe.

But it's still not trivial to follow, hence my not entirely confident statement.

Beyond fixing is_ansi_end, I'd recommend to increase the buffer to 8K to (greatly) reduce the likelyhood of needing to handle EOB, at least on windows or maybe all the MSDOS_COMPILER.

There's also the issue of overlapped `strcpy` which can only happen with an esc sequence bigger than half the buffer size, and still depending on implementaqtion copying order, but let's fix it anyway.

> there's indeed a bug that if the last char at the buffer is ESC, then it's not considered incomplete sequence

This indeed exists, The implication is that the ESC would be flushed directly instead of interpreted by the SGR processor. The bug is largely not visible on win10+ because the terminal can handle the esc sequence in such case (but we do lose sync of the SGR state).

It's easy to fix, but don't think we should touch it because it's non critical and ulikely to rear its head in practice, especially if we increase the buffer size (needs line long enough to spill the buffer AND a sequence which begins at the last byte of the buffer).

Bottom line:
- Fix `is_ansi_end` in this function only.
- Replace the `strcpy` with `memmove`,
- Increase `obuf` from 1K to 8K with MSDOS_COMPILER to decrease likelyhood of EOB bugs, of which we're currently aware of one non-critical issue.

@gwsw agreed? anything more or less? shall I update this PR to these things?

avih:
Ping? So how do you want to handle this?

gwsw:
Sorry, I didn't see that you had asked a question in your second to last post. I agree with the three proposals you made at the end of that post. Go ahead and update the PR with those changes.

avih:
> Go ahead and update the PR with those changes.

Done.

gwsw:
Ironically, I just found another bug caused by is_ansi_end(0) returning TRUE, which is entirely unrelated to this issue (#550). But it's not a regression, having existed in the past 7 major releases, so I don't plan to address it in the upcoming bug fix release.

avih:
> Ironically, I just found another bug caused by is_ansi_end(0)

Yeah. I'm pretty sure nothing expects it to be 0. Whether it ends up buggy or not depends on circumstances.

But yeah, I agree it's best not change it globally for the bugfix release.

avih:
Please wait with the release.

I've discovered another related issue where with `-Da` it can now exceet the `WIN32textout` internal buffer.

I'll PR a fix soon.

avih:
> I'll PR a fix soon.

Filed  #562 .

gwsw:
@avih, are you still planning to submit a rewrite of the Windows SGR parser? If so, do you have an idea of when you will do that?

avih:
I want to, but I don't know when, and I don't have a plan, so don't wait for me.

I need to sit down and decide how to make true colors work out of the box on win10 (when VT is enabled but without `-Da`), because that's what v668 (and current master) does, but the parser was written before we added this behavior, and it's not immediately compatible with it, and so far I haven't done that.

PR #567, opened 2024-09-07
Allow to override build date with SOURCE_DATE_EPOCH

bmwiedemann:
Allow to override build date with `SOURCE_DATE_EPOCH` to make builds reproducible.
See https://reproducible-builds.org/ for why this is good and https://reproducible-builds.org/specs/source-date-epoch/ for the definition of this variable.


gwsw:
I'm not sure I understand the purpose of this. This patch changes only a **comment** in the generated help.c file. It does not change anything in the generated binaries. As far as I can tell, nothing in the less binaries changes based on the build date. If you see a dependency on the build date can you explain where you see it?

bmwiedemann:
The diff was in our less-debugsource rpm package that is used by `gdb` to show source code.

gwsw:
If you care about the source itself being reproducible, wouldn't you also want to munge the dates in the .uni files?

bmwiedemann:
This patch was all it needed for bit-reproducible results in openSUSE. I guess, the debugsource packages only contain .h and .c files, but not .uni files.

gwsw:
Implemented in c02f7554a1e5685e4332fb6857e95761953c8db4 which for consistency also uses SOURCE_DATE_EPOCH when generating the .uni files.

Issue #152, opened 2021-04-17
Have an option to show tabs as tabs


guillaumebrunerie:
It would be nice to have an option to get `less` to show tabs as tabs (not convert them to spaces), so that for instance selecting text in the terminal window and pasting it into another program will include actual tab characters.

I found the following Stack Exchange question (and answer) https://unix.stackexchange.com/questions/412060/how-to-get-less-to-show-tabs-as-tabs, which suggest that it is currently impossible, but that it would be easy to change.

gwsw:
I don't see how it would be possible to do this and still break lines correctly. Without knowing how many spaces the terminal uses for a tab, less wouldn't know how long the line is, so wouldn't know where the line breaks are or how many lines fit on a screen.

gwsw:
Well I guess less could set hardware tabs to match the desired tabs set via -x, but this seems fairly nontrivial.

busterbeam:
I think it should use tabs.  And wrap assuming 8 tabstop as that is what less default changes tabs too?

gwsw:
Repeating my comment from #292 for reference:

It would be possible, but rather challenging to do this. For example, if the text had long lines which wrapped, then depending on the relationship between the tabstop settings and the width of the terminal, less would need to set different terminal tabs before printing every single line. Also if the text were scrolled horizontally with the arrow keys, tabs would need to be set to a "shifted" version of the -x settings. And I don't believe there's any way to read tab settings from a terminal, so when less exits there would be no way to restore tab settings to what they were before less was run. I notice that vim also uses spaces rather than hard tabs, presumably for similar reasons.

busterbeam:
@gwsw if you read my comment
> wrap assuming 8 tabstop

as `less` (if not otherwise specified) assumes 8 spaces for tabs already, so use an actual tab character and assume it is 8 spaces.  Then end user can setup the rest.

in fact some research... I think you should read into `tabs` command in linux/bash not unlike the `clear` command.  and then use `reset` to return to initial conditions... have you ever read into any of this?  I will admit I have not explored this repository much so I don't know exactly how it is all implemented but I imagine real tabs can be implemented

gwsw:
I don't think you understood my previous comments. Yes, I understand that terminal tabs can be set using the "st" termcap capability ("set_tab" in terminfo) as the tabs command does, but that's only a small part of the issue.

[...]

PR #620, opened 2025-04-30
Fix tabs not preserved in raw mode

mohd-akram:


gwsw:
Can you explain the motivation for this change? Normally less does not output tabs, but can be configured to do so via the -U option. In that case, currently less configures the tty driver to expand tabs to spaces. This PR configures the driver to output raw tabs instead. I think an argument could be made for either behavior. I can also see an argument to leave the TAB bits unchanged so that the tty driver does whatever the user has configured it to do outside less.

mohd-akram:
The primary motivation is to be able to copy diffs when eg. using `less` as a pager for git (#152).

> I think an argument could be made for either behavior.

IMO this behavior is more useful because the current behavior is a no-op for tabs in raw mode, and if one still wants expansion in raw mode, they can get it with `--proc-tab`. I also noticed the following comment, so it seems this was the intention:

https://github.com/gwsw/less/blob/26cfc730421dc017bade728df2eab43a5de20087/screen.c#L373

> I can also see an argument to leave the TAB bits unchanged so that the tty driver does whatever the user has configured it to do outside less.

That's what my patch did initially, but then I saw:

https://github.com/gwsw/less/blob/26cfc730421dc017bade728df2eab43a5de20087/screen.c#L631

I guess that line would also need to be tweaked? I don't have a strong opinion on this part tbh, and since the user can control it, it doesn't hurt.

gwsw:
I think the modified patch is ok. It was clearly wrong for the behavior to be different depending on whether TCGETA or TIOCGETP is supported. 

However, I think the change should be made only to screen.c, not subsearch.lt nor term.c. The subsearch.lt file is not source code, it is just using an old copy of screen.c as a text file for one of the tests. And in term.c which is part of the lesstest program, the behavior should be deterministic, not dependent on how the user has set his terminal. The exact behavior (tabs expanded or not) doesn't really matter, but it does matter that the behavior is the same for every user, regardless of their tty configuration.

PR #622, opened 2025-05-15
WIN32: edit guide to build for MSVC.

groutoutlook:
There are missing guide on how to build in Windows environment using MSVC


avih:
i can confirm the additional instructions for building using MSVC compiler are correct (they're copied from `Makefile.wnm`).

No idea about the additional info about Borland compiler, and I suspect it's incorrect - unless OP can demonstrate otherwise.

groutoutlook:
I don't have a sttock Borland C Compiler either. Where should I progress?

groutoutlook:
Tunnel vision, don't know why I only see Borland CC... At least must have stated "Otherwise" instead.

avih:
Actually, the additional instructions conflict with step 2 - which was not modified.

Step 2 instructs to rename `Makefile.wnm` to `Makefile`, and therefore `Makefile.wnm` would not exist anymore.

Also, the first step of the new instructions (generating help.c and funcs.h) is only required for git snapshots - not for release tarballs.

So basically the instructions for release tarballs were almost OK (I think it needs `nmake` instead of `make`) and now they conflict with step 1.

As for borland, as I said, I don't think it would build. It might be possible to make it build, but no one uses Borland anymore, and I wouldn't know how to setup a Borland build env to test that.

Also, notably missing is the mingw build instructions, which these days should be the prominent environment, because it's open source - unlike MSVC.

So not sure what to say overall, but it needs to ensure all the steps are at least consistent with eachother, and in this PR they're not consistent.

groutoutlook:
I have `make` install in my system, it couldn't run `make all` as well as `nmake` for some reason... 

In case of instructions, I have updated it, please take a look

avih:
> I have `make` installed in my system, it couldn't run `make all` as well as `nmake` for some reason...

I don't understand what this sentence means.

While I know MSVC ships with `nmake`, I don't know whether it also has `make`.

Is `make` part of the MSVC package? or is it installed independently? (where from?).

What does this mean? "as well as `nmake` for some reason" - that running `nmake` (instead of `nmake all`) after renaming `Makefile.wnm` to `Makefile` didn't work? That doesn't sound reasonable to me...

Otherwise, the changes at README look OK, but the changes at `.gitignore` are missing a final newline.

groutoutlook:
About Mingw-w64, `make` is working exactly as intended. Although we dont have a `generated` rule there.

groutoutlook:
yeah I will leave that there. I just dont want to change his wording too much so you know..

Issue #627, opened 2025-05-20
Feature Request: Supporting multiple actions in lesskey file


druckdev:
I'd love to be able to bind multiple actions to a key in the lesskey file. Specifically I'm asking, since I want to bind the action `goto-line` followed by `pipe $vim -\n`.

Since this would require changes to the file format, I can imagine that it might break backwards-compatibility. An alternative to this could be a new action, that parses and executes the extra-string as if it was typed into `less` (Similar to how the right-hand-side of vim's `map` command works). For my use-case one could then bind `<new-action> g|$vim -\n`

Is something like this imaginable?
Thanks!

gwsw:
I don't think I am understanding this request. You can put any characters in the action field of a lesskey binding, and that command acts as if the action string were typed into less. For example, if you wanted to make a ```K``` command that does goto-line followed by a pipe of the current screen into vim, you could use something like
```
K goto-line |.vim -\n
```
Are you asking for something that can't be done in this way?

druckdev:
Ah very interesting. I was misled a bit by the manpage that I misread as if one could only bind one action per key(s), but have to admit that I didn't think about using the extra-string for an action that does not need it (unlike for example `forw-bracket`).

I have the following in my lesskey now:

```
v noaction g|$vim -\n
```

Thanks a lot!

PR #630, opened 2025-05-31
lesskey: Give an example for multi-action bindings in the manpage

druckdev:
Since I falsely assumed that one could bind only one action per key (#627), I think that it could be helpful also to others to explicitly explain how the extra-string can be used to bind multiple actions.

This also fixes a small indentation issue in the same section.


gwsw:
I made a similar change in 913eb926cdf8f34ae5612f801615e594083d6962 but with a simpler example.

Issue #654, opened 2025-09-12
Bug: flickering on horizontal scroll


rszyma:
Screen flickers on horizontal scroll in less.

Steps: 
1. Run `ls -la /proc | less -R -# 2` in alacritty
2. Press right or left arrows and observe flickering

Video:

https://github.com/user-attachments/assets/21ec6889-08e6-4106-be9f-e8626c235ae3

I can reproduce in following terminals: alacritty, gnome-terminal.
Can't reproduce in following terminals: kitty, foot, konsole, st, xterm, zed built-in terminal, vscode built-in terminal.

I'm not exactly sure exactly whether the problem is within alacritty/gnome-terminal or less not doing proper sync. One of alacritty devs in [this comment](https://github.com/alacritty/alacritty/issues/6709#issuecomment-1438311618) claims that it's an issue on less side.

I've done some grepping around less' code searching for any mentions of sync escape sequences, but I couldn't find anything relevant, which makes me more confident it's indeed an issue on less side.

avih:
That's how applications render a new "image" to the terminal: clear the screen and print something else. It's an expected behavior with applications which need to update the whole screen, depending on how the application updates it, and generally should not be considered an issue.

Depending on implementation, timing, and heuristics, terminals may or may not draw the in-between clear screen state, or maybe only sometimes render it.

Historically, there's no way for the application to tell the terminal "I'm now printing new content and you should not update the visible screen until I tell you I'm done with the update".

In recent years there have been some suggestions about "synchronized updates", where the application can instruct the terminal to suspend visible screen updates until the application tells it to continue, and it is supported by some terminals, but there's no standard for that, as far as I know, in termcap or terminfo, though tmux did come up with something ("Sync" capability), but I don't know how many terminals export it.

Less doesn't use syncronized updates, so it's up to the terminal to decide at which points in time to render the screen, and if that happens to be after the clear but before the new content is printed, then it's perceived as flicker.

gwsw:
[...]

PR #655, opened 2025-09-14
Universal fix for flicker on horizontal scroll.

rszyma:
This, somehow, significantly reduces flickering on horizontal scroll in flicker-sensitive terminals (like alacritty),
without needing to set SUSPEND and RESUME sequences (added in 77dddb2).

Separately, this also (mostly) prevents rendering (and flickering) of string representation printed in command line when handling left/right keys.

Related to #654.

### DIFF
--- a/command.c
+++ b/command.c
@@ -2288,6 +2288,7 @@ public void commands(void)
 			pos_rehead();
 			hshift -= (int) number;
 			screen_trashed();
+			cmd_exec();
 			break;
 
 		case A_RSHIFT:
@@ -2301,6 +2302,7 @@ public void commands(void)
 			pos_rehead();
 			hshift += (int) number;
 			screen_trashed();
+			cmd_exec();
 			break;
 
 		case A_LLSHIFT:
@@ -2310,6 +2312,7 @@ public void commands(void)
 			pos_rehead();
 			hshift = 0;
 			screen_trashed();
+			cmd_exec();
 			break;
 
 		case A_RRSHIFT:
@@ -2319,6 +2322,7 @@ public void commands(void)
 			pos_rehead();
 			hshift = rrshift();
 			screen_trashed();
+			cmd_exec();
 			break;
 
 		case A_PREFIX:


rszyma:
To make it clear, this is not a full replacement for 77dddb2024f716e6ab64ff3894ec2dae310130fd, rather a general remedy. With this patch, and without the sync sequences set, I still can observe some flickering but much less frequent, but only if I zoom out terminal very far, but on normal zoom I practically no longer see any flickering (of the file content).

gwsw:
Hm, I don't see much difference in flickering in my testing with an alacritty terminal. Calling cmd_exec() after calling screen_trashed() shouldn't make a significant difference, although it causes writes to the terminal to be done at different points, which may have a side effect of reducing flickering depending on the terminal's update algorithm, the screen size or even the file contents. I'm reluctant to add this seemingly unnecessary change without a better understanding of why it affects flickering, but I'll do some investigation. Can you provide the screen size and the data file that you are using in your testing?

gwsw:
Also, just to help me understand what's happening, do you see the same improvement if you replace the calls to ```cmd_exec()``` with ```flush()``` at  those same four points?

rszyma:
> Also, just to help me understand what's happening, do you see the same improvement if you replace the calls to cmd_exec() with flush() at those same four points?

I do! When I replace `cmd_exec()` with `flush()` it reduces flickering of the screen, but it keeps flashing the arrow key string code in command line. And when I replace `cmd_exec()` with `clear_bot(); flush();` it additionally prevents the displaying arrow key repr, so `clear_bot` is useful to keep. The other left part is `clear_attn()`, it appears that it's not needed, but I'm not really fully understanding when exactly it needs to be called to tell that for sure.

> Can you provide the screen size and the data file that you are using in your testing?

It's obviously going to be different from PC to PC, how many chars on screen terminal can handle unsynced until substantial flashing starts appearing. For the following setup I'm only getting some flashes, and for me it's much less likely to occur than without the extra `flush()`:

1920x1080@180hz, alacritty: fullscreen, zoomed out as much as possible, keyboard repeat rate 50 ms, Source file + command: `cat /dev/urandom | tr -dc '!-~' | fold -w 150 | ./less-debug-build -# 2` - roughly on average getting 1 flash every 5 seconds or so (with this patch and without sync sequences). Without the patch and without sync sequences, was multiple times a second, so much much worse, I've shown on the video in https://github.com/gwsw/less/issues/654.

avih:
It's worthwhile to remember that while alacritty might be a good test case for synchronize updates, it doesn't represent anything but itself regarding this PR.

I.e. the fact that this PR might make it work better in alacritty doesn't mean it would also improve in other terminals, because ultimately it depends on terminal implementation details and when it chooses to update the visible screen, and this differ between terminals.

Further than that, IIRC alacritty specifically is different than many other terminals approach, so without additional testing with many terminals, it should not be assumed that this PR improves anything other than alacritty.

Other than synchronized updates, there's only one thing which is generally useful among different terminals, and that's to minimize the time where the bad screen state stays before further output is sent so that it's no longer consider an undesirable state, for instance, reduce the time between clear screen and additional output which fills the screen.

So I think this should be kept in mind.

gwsw:
I've done some investigation and I can't see why this would affect flickering. The only thing that gets flushed when the new flush() calls are done is writing the arrow escape sequence to the command line. I suspect that the changed sequence of writes to the terminal is triggering different alacritty behavior in some way that is not easily described without studying the alacritty internals. As avih suggests, this may not behave the same in other terminals.

Nevertheless, I don't see that this change would cause any harm, and the clear_bot() does potentially reduce the amount of time that the undesirable arrow escape sequence is displayed, so I will merge it.

PR #656, opened 2025-09-16
Add z/OS compatibility improvements

IgorTodorovskiIBM:
This PR adds compatibility improvements to make the `less` utility work properly on z/OS systems. The changes address three specific issues that prevent proper operation on this platform.

## Changes
**UTF-8 charset support**: Added code to default to UTF-8 charset on z/OS systems since the z/OS locale implementation doesn't properly support UTF-8
**SIGPIPE handling**: Modified signal handling to conditionally check for SIGPIPE only when the signal is defined, as some systems (z/OS) may not define this signal
**poll() functionality**: Excluded z/OS systems from using poll which causes issues when going page up or down.

If you'd like to test these changes or improve upon them for z/OS, please let me know and I can get you access to a z/OS system.

### DIFF
--- a/charset.c
+++ b/charset.c
@@ -456,6 +456,12 @@ static void set_charset(void)
 	 * rather than from predefined charset entry.
 	 */
 	ilocale();
+#ifdef __MVS__
+/*
+	 * z/OS Locale does not properly support UTF-8. Default to "utf-8".
+*/
+  (void) icharset("utf-8", 1);
+#endif
 #else
 #if MSDOS_COMPILER
 #if MSDOS_COMPILER==WIN32C
--- a/os.c
+++ b/os.c
@@ -43,7 +43,7 @@ extern int errno;
 #include <sys/utsname.h>
 #endif
 
-#if HAVE_POLL && !MSDOS_COMPILER
+#if HAVE_POLL && !MSDOS_COMPILER && !defined(__MVS__)
 #define USE_POLL 1
 static lbool use_poll = TRUE;
 #else
--- a/edit.c
+++ b/edit.c
@@ -306,7 +306,11 @@ static void close_pipe(FILE *pipefd)
 	if (WIFSIGNALED(status))
 	{
 		int sig = WTERMSIG(status);
-		if (sig != SIGPIPE || ch_length() != NULL_POSITION)
+		if (
+#ifdef SIGPIPE
+      sig != SIGPIPE ||
+#endif
+      ch_length() != NULL_POSITION)
 		{
 			parg.p_string = signal_message(sig);
 			error("Input preprocessor terminated: %s", &parg);


charset.c
gwsw, line 1: Why is this necessary? The existing code calls ```icharset("utf-8",1)``` unless MSDOS_COMPILER is not zero and not equal to WIN32C. How does the z/OS build set MSDOS_COMPILER?

os.c
gwsw, line 1: Could the z/OS build just set HAVE_POLL to 0, to avoid the need for an extra condition here?

avih:
It would be useful to know how exactly you're building "less", as it can shed some light on macros/flags which might or might not be auto-detected, etc.

os.c
IgorTodorovskiIBM: Do you recommend that we just add this to the CFLAGS? -DHAVE_POLL=0?

IgorTodorovskiIBM:
> It would be useful to know how exactly you're building "less", as it can shed some light on macros/flags which might or might not be auto-detected, etc.

Our build scripts are a little complex. We use a tool called `zopen-build` that automates the build for various OSS tools but in general the default options that we pass are in https://github.com/zopencommunity/check_clangport/blob/main/buildenv#L103 and https://github.com/zopencommunity/check_clangport/blob/main/buildenv#L14

gwsw:
Which defines.h do you use? Do you use one of the checked-in versions, or do you create your own by running configure or by some other process?

avih:
> Our build scripts are a little complex. We use a tool called `zopen-build` that automates the build for various OSS tools but in general the default options that we pass are in https://github.com/zopencommunity/check_clangport/blob/main/buildenv#L103 and https://github.com/zopencommunity/check_clangport/blob/main/buildenv#L14

I fail to see where it actually does the build. I see it sets
```sh
export ZOPEN_CONFIGURE=skip
export ZOPEN_MAKE=skip
```

which supposedly skips configure and make? if it does use make, which makefile is used? where is the actual invocation of the compiler? where's the list of files to compile and link?

IgorTodorovskiIBM:
> Which defines.h do you use? Do you use one of the checked-in versions, or do you create your own by running configure or by some other process?

We run configure via the buildenv configuration - https://github.com/zopencommunity/lessport/blob/main/buildenv (zopen-build will run configure by default unless it's skipped)



> > Our build scripts are a little complex. We use a tool called `zopen-build` that automates the build for various OSS tools but in general the default options that we pass are in https://github.com/zopencommunity/check_clangport/blob/main/buildenv#L103 and https://github.com/zopencommunity/check_clangport/blob/main/buildenv#L14
> 
> I fail to see where it actually does the build. I see it sets
> 
> ```shell
> export ZOPEN_CONFIGURE=skip
> export ZOPEN_MAKE=skip
> ```
> 
> which supposedly skips configure and make? if it does use make, which makefile is used? where is the actual invocation of the compiler? where's the list of files to compile and link?

You can ignore those - I'm just pointing out the default cflags that we pass via the command line.

A sample compile invocation looks as like this:
```
clang -I. -c -DBINDIR=\"/home/itodoro/zopen/usr/local/zopen/less/less-master/bin\" -DSYSDIR=\"/home/itodoro/zopen/usr/local/zopen/less/less-master/etc\" -DNSIG=42 -D_XOPEN_SOURCE=600 -D_ALL_SOURCE -D_OPEN_SYS_FILE_EXT=1 -D_AE_BIMODAL=1 -D_ENHANCED_ASCII_EXT=0xFFFFFFFF -DLESSTEST=1 -I/home/itodoro/zopen/usr/local/zopen/ncurses/ncurses-6.5.20250625_114034.zos/include/ncurses -DZOSLIB_OVERRIDE_CLIB=1 -DZOSLIB_OVERRIDE_CLIB_GETENV=1 -DZOSLIB_OVERRIDE_CLIB_LOCALE_FORCE=0 -DZOSLIB_OVERRIDE_CLIB=1 -DZOSLIB_OVERRIDE_CLIB_GETENV=1 -DZOSLIB_OVERRIDE_CLIB_LOCALE_FORCE=0 -std=gnu11 -fzos-le-char-mode=ascii -mnocsect -fno-short-enums -m64 -mzos-target=zosv2r5 -march=z13 -fno-builtin-malloc -fno-builtin-calloc -fno-builtin-realloc -O3 -mzos-target=zosv2r5 -march=z13 -DPATH_MAX=1024 -isystem /home/itodoro/zopen/usr/local/zopen/zoslib/zoslib-zopen2.20250825_142938.zos/include -include /home/itodoro/zopen/usr/local/zopen/zoslib/zoslib-zopen2.20250825_142938.zos/include/zos-v2r5-symbolfixes.h  lsystem.c
```

avih:
So it first runs `make -f Makefile.aut help.c defines.h.in funcs.h distfiles`, and autoconf something, and then `./configure` and then `make` ?

IgorTodorovskiIBM:
> So it first runs `make -f Makefile.aut help.c defines.h.in funcs.h distfiles`, and autoconf something, and then `./configure` and then `make` ?

Yep, that is correct

avih:
Can you post the output of `./configure`?

On linux it prints `checking for poll... yes`, so I presume on z/OS it's `no`? Shouldn't this disable `HAVE_POLL`, and make some of the changes in this PR redundant?

I'll let @gwsw continue, I gotta go for now.

IgorTodorovskiIBM:
> Can you post the output of `./configure`?
> 
> On linux it prints `checking for poll... yes`, so I presume on z/OS it's `no`? Shouldn't this disable `HAVE_POLL`, and make some of the changes in this PR redundant?
> 
> I'll let @gwsw continue, I gotta go for now.

Yes, poll was detected:
```
checking for poll... yes
```
However, the poll implementation on z/OS has issues when paging up or down or when searching for text (likely some z/OS bug). That's why I disabled it in the code.

gwsw:
Ok, we do have some special cases in the code to disable poll on a couple of systems (MSDOS & Windows, and some versions of MacOS). I think it's ok to add another one for z/OS but let me think about it a little and see if there might be a better way to handle this. Maybe an option to configure would be better.

The change for SIGPIPE looks ok.

For the utf-8 issue, could you set LESSCHARSET=utf-8 in your system environment? That should override the locale setting and wouldn't require a change to the code.

avih:
Before the final comit is merged, it's worth explaining at the commit message why it does things the way it does, for instance, it should include something like "./configure on z/OS does detect that poll is supported, however, it's buggy on z/OS version XXX and earlier, so this commit disabled it explicitly", and similarly the reaons for any other changes which are not obvious.

Maybe some comments at the code too where something is not obvious.

Otherwise in the future it would not be possible to tell why HAVE_POLL is bypassed on z/OS, etc, and someone might remove this code because there's no apparent reason to bypass it.

Or maybe the poll bug is fixed in z/OS at some stage, so the exception can be removed at the code (maybe), but it will never happen because no one could tell why it's there in the first place, so it becomes zombie code which no one understands why it's there.

gwsw:
@IgorTodorovskiIBM I have implemented two of your three changes. Please let me know if it's possible for you to set LESSCHARSET=utf-8 in your environment to solve the third problem as I suggested above, or if there's some reason that that wouldn't work for you.

IgorTodorovskiIBM:
> @IgorTodorovskiIBM I have implemented two of your three changes. Please let me know if it's possible for you to set LESSCHARSET=utf-8 in your environment to solve the third problem as I suggested above, or if there's some reason that that wouldn't work for you.

Hi @gwsw yes that works. Thank you!

gwsw:
Changes implemented in 84acbf7d783c2d879b68624ffcdae36460dba16d and 375e3d934c6f0ec7a5e598ae9b6e25b0d744c245.

PR #663, opened 2025-10-12
Revise man pages

g-branden-robinson:
Address style issues and portability to AT&T _troff_ systems such as DWB 3.3, Solaris 10, and Plan 9.  (Heirloom Doctools _troff_ is also descended from AT&T _troff_, but has incorporated many _groff_ extensions).


gwsw:
I had to do a little manual merging because the man pages were changed after you submitted this PR. Let me know if you see any problems.

PR #669, opened 2025-10-21
Less gbr man page changes

g-branden-robinson:
Improve portability and robustness of _less_'s man pages.

### DIFF
--- a/less.nro.VER
+++ b/less.nro.VER
@@ -254,7 +254,7 @@ the status column shows the marked line.
 .IP M
 Acts like m, except the last displayed line is marked
 rather than the first displayed line.
-.IP "\(aq"
+.IP "\*'"
 (Single quote.)
 Followed by any lowercase or uppercase letter, returns to the position which
 was previously marked with that letter.
@@ -264,7 +264,7 @@ Followed by a \*^ or $, jumps to the beginning or end of the
 file respectively.
 Marks are preserved when a new file is examined within a single invocation of
 .BR less ,
-so the \(aq command can be used to switch between input files.
+so the \*' command can be used to switch between input files.
 The \-\-save\-marks option causes marks to be preserved across different invocations of
 .BR less .
 .IP "\*^X\*^X"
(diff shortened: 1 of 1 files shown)


gwsw:
Is every hyphen supposed to be escaped? I notice that even after applying this PR, this command returns 169 lines:
```
$ grep '[^\\]-' *.nro.VER
```

g-branden-robinson:
> Is every hyphen supposed to be escaped?

No.  That's one of the reasons _groff_ doesn't just remap `-` to `\-` itself.

_groff_man_style_(7):
```
     • Some ASCII characters look funny or copy and paste wrong.

       On devices with large glyph repertoires, like UTF‐8‐capable
       terminals and PDF, GNU troff, like AT&T troff before it, maps
       several keycaps to code points outside the Unicode basic Latin
       range (historically “ASCII”) because that usually results in
       better typography in the general case.  When documenting
       GNU/Linux command or C language syntax, however, this translation
       is sometimes not desirable.

       To get a “literal”...   ...should be input.
       ────────────────────────────────────────────
                           '   \[aq]
                           -   \-
                           \   \[rs]
                           ^   \[ha]
                           `   \[ga]
                           ~   \[ti]
       ────────────────────────────────────────────

       If a neutral double quote (") is needed in a macro argument, you
       can use \[dq] to get it.  Do not use \[aq] for an ordinary
       apostrophe (as in “can’t”) or \- for an ordinary hyphen (as in
       “word‐aligned”).
```

(The foregoing advice assumes use of "modern" formatters like _groff_, _mandoc_, or Heirloom Doctools _troff_.   My recommendations to projects like _less_ where portability to legacy Unix is desired are a little different.)

_groff_char_(7) has more information.

``` 
     The hyphen‐minus is a particularly unfortunate case of overloading.
     Its awkward name in ISO 8859 and later standards reflects the many
     distinguishable purposes to which it had already been put by the
     1980s, including a hyphen, a minus sign, and (alone or in
     repetition) dashes of varying widths.  For best results in roff
     systems, use the “-” character in input outside an escape sequence
     only to mean a hyphen, as in the phrase “long‐term”.  For a minus
     sign in running text or a Unix file name or command‐line option
     dash, use \- (or \[-] in groff if you find it helps the clarity of
     the source document).  (Another minus sign, for use in mathematical
     expressions, is available as \(mi.)  AT&T troff supported em‐dashes
     as \(em, as does groff.
```

> I notice that even after applying this PR, this command returns 169 lines:
> 
> ```
> $ grep '[^\\]-' *.nro.VER
> ```

I used the same search pattern and left intact every unescaped hyphen that looked contextually okay to me.

PR #706, opened 2026-01-03
Mingw refinements

avih:
Two `Makefile.wng` minor changes: one only refines the regex comment, and the other adds some build flags options without changing the default behavior.


avih:
Thanks.

PR #714, opened 2026-01-25
Fixed support of old Perl interpreters

polluks:
At least version 5.8 works fine.


avih:
Just FYI, we have [mkhelp.sh](https://github.com/gwsw/less/blob/master/mkhelp.sh) too, which only requires sh, awk, and od, all of which are standard POSIX.

The mingw makefile already falls back to it if mkhelp.pl fails to run (typically due to missing perl), so maybe the unix makefile could fall back to it too.

PR #765, opened 2026-05-01
man page fixes and improvements

g-branden-robinson:
Most of these commits are (conceptually) independent of the others.  Let me know if you'd like me to excise any of them.


gwsw:
I will do a more thorough evaluation soon. At first glance, my biggest concern is the mixed case section headings. This seems quite anomalous compared to other man pages. On my Fedora 6.17 system, there are 10,375 man pages. Only 162 of them use mixed case section headings, and all of these belong to just two packages: groff and nodejs. There may be reasonable arguments for why mixed case headings are in some ways "better", but I would not like less's man page to be perceived as an outlier or doing something non-standard.

g-branden-robinson:
> At first glance, my biggest concern is the mixed case section headings. This seems quite anomalous compared to other man pages.

I concede--that commit is the most "reformist" of the batch by a considerable margin.

> Only 162 of them use mixed case section headings, and all of these belong to just two packages: groff and nodejs.

A stranger bedfellow than I would ever have imagined!

> There may be reasonable arguments for why mixed case headings are in some ways "better", but I would not like less's man page to be perceived as an outlier or doing something non-standard.

Okay.  Happy to drop and re-push if desired.  Let me know.

gwsw:
Ok, please remove the mixed case heading change and repush, and I'll review the revised PR.

g-branden-robinson:
Updated with some clarifications and corrections to the commit messages.  The last was a real howler: swapped font names characterized the problem as the fix.

gwsw:
Perhaps I misunderstood the previous conversation about mixed-case section headings. I still see such headings in the PR but I thought that you were going to leave them as all uppercase. For example, I see "Commands" not "COMMANDS", "Line editing" not "LINE EDITING", "Key bindings" not "KEY BINDINGS", etc.

g-branden-robinson:
> Perhaps I misunderstood the previous conversation about mixed-case section headings. I still see such headings in the PR but I thought that you were going to leave them as all uppercase. For example, I see "Commands" not "COMMANDS", "Line editing" not "LINE EDITING", "Key bindings" not "KEY BINDINGS", etc.

Ah, yes, probably we miscommunicated.

Those headings are now **sub**sections.  That's what this stuff is about:

```
The `SS` macro is universally portable, dating back to Seventh Edition
Unix (1979)--albeit undocumented in that release's own man(7) page.

Use sentence case for subsection headings.  While existing practice is
inconsistent, and in documents originating post-Linux-kernel-era (1991)
wildly so, mixed case is preponderantly seen in historical Unix man
pages.  Admittedly, that same perponderance also (mostly) employs title
case, not sentence case.  However, mandoc(1) maintainer Ingo Schwarze
and I both feel that `SH` section headings should be set in mixed case
as well, because it is reliable to convert them to full capitals at
rendering time if desired.  Reportedly, some people using screen readers
do _not_ desire full capitals, because they might spelled out letter by
letter like an initialism ("en ay em ee", cf. "name").

References:

https://minnie.tuhs.org/cgi-bin/utree.pl?file=V7/usr/lib/tmac/tmac.an
https://paste.c-net.org/DetailsStalker
https://lists.gnu.org/archive/html/groff/2019-07/msg00026.html
https://lists.gnu.org/archive/html/groff/2020-09/msg00016.html
```

What's your preference?  Kick these back up to "full" sections and therefore fully capitalize them?  Keep them as subsection headings but _also_ fully capitalize them, or leave the PR as-is in this respect?

gwsw:
Oh, I see. I guess I don't have a strong objection to subdividing the manual into subsections in principle. However the way it's divided in this PR seems odd to me. For example, under the OPTIONS section, we have subsections "Line editing", "Key bindings", "Input preprocessor", "National character sets", etc., none of which IMHO have anything to do with OPTIONS. Looking at the existing set of section headings, I'm not sure I see any logical way to move any of them into subsections. They all seem fairly independent of each other. So I guess I'd prefer to leave them all as top-level sections (.SH), unless you think that I'm missing something.

g-branden-robinson:
> Oh, I see. I guess I don't have a strong objection to subdividing the manual into subsections in principle. However the way it's divided in this PR seems odd to me. For example, under the OPTIONS section, we have subsections "Line editing", "Key bindings", "Input preprocessor", "National character sets", etc., none of which IMHO have anything to do with OPTIONS. Looking at the existing set of section headings, I'm not sure I see any logical way to move any of them into subsections. They all seem fairly independent of each other. So I guess I'd prefer to leave them all as top-level sections (.SH), unless you think that I'm missing something.

Nope, I don't think so.  Over time maybe a more subsectioning-friendly arrangement of material can be hammered out.

I'll back out that commit, too, resolve any merge conflicts, and re-push.

gwsw:
BTW, it would be better to base this PR on the post700 branch rather than master.

g-branden-robinson:
Rebasing onto `post700` made a mess of this.  Closing.

PR #767, opened 2026-05-05
man page fixes and improvements

g-branden-robinson:
Most of these commits are (conceptually) independent of the others. Let me know if you'd like me to excise any of them.

Rebased on `post700`.  Updated version of #765.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -40,7 +40,8 @@ sysconfdir = @sysconfdir@
 datarootdir = @datarootdir@
 
 mandir = @mandir@
-manext = 1
+manprogext = 1
+manfileext = 5
 manprefix = 
 
 #### End of system configuration section. ####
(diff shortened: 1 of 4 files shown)


avih:
This is only an estimation, but if there are docs changes/conflicts between current master and the post700 branch, then maybe it would be better to wait for it to be merged into master, and then post the docs changed PR on top of the new master?

Because otherwise (more docs changes to post700), I think it would make the following master-merge more complex?

g-branden-robinson:
> This is only an estimation, but if there are docs changes/conflicts between current master and the post700 branch, then maybe it would be better to wait for it to be merged into master, and then post the docs changed PR on top of the new master?
> 
> Because otherwise (more docs changes to post700), I think it would make the following master-merge more complex?

I'm doing [what Mark explicitly asked me to do](https://github.com/gwsw/less/pull/765#issuecomment-4365705360).

avih:
> I'm doing [what Mark explicitly asked me to do](https://github.com/gwsw/less/pull/765#issuecomment-4365705360).

Then this was for him too ;)

PR #780, opened 2026-05-26
Fixes #779: Argument injection less osc8

### LINKED ISSUE #779: Argument injection in less-osc8-open enables RCE on link activation
The shipped script `less-osc8-open.sh` is registered as the default value of `LESS_OSC8_OPEN_ANY` in `decode.c`:

```c
static char dflt_vartable[] =
{
    "LESS_OSC8_OPEN_ANY\0\201"
#ifdef LIBEXECDIR
        "-" LIBEXECDIR "/less-osc8-open"
#else
        "-less-osc8-open"
#endif
        "\0"
};
```

When the user activates an OSC 8 hyperlink (Ctrl-O Ctrl-O, or a second mouse click in mouse mode), `osc8_open()` in `search.c` invokes the handler with the URI as a single shell-quoted argument:

```c
uri_q = shell_quoten(op.uri_start, uri_len);
cmd = ecalloc(strlen(handler) + strlen(uri_q) + 2, sizeof(char));
sprintf(cmd, "%s %s", handler, uri_q);
...
lsystem(exec_cmd, done_msg);
```

That delivers the URI verbatim into the script as `$1`. Inside the script, the `man:NAME(SECTION)` case splits the URI:

```sh
case $1 in
man:?*\(*\) )
    sect=${1#*\(}; sect=${sect%?}
    name=${1#man:};   name=${name%%\(*}
    man ${sect:+"$sect"} "$name"
    ;;
```

`${sect:+"$sect"}` expands to a single argv word, but `man` still interprets a word beginning with `-` as an option. Both GNU `man-db` and BSD `man` accept `-P pager`/`-Ppager`/`--pager=pager` with shell-tokenised pager strings, and `man-db` documents that the value "may be a simple command name or a command with arguments." Tokenisation alone is sufficient for execution: the pager argv is `execvp`'d with man's formatted output on stdin.

### Reproduction

```text
URI inside OSC 8 sequence:    man:ls(-P /usr/bin/touch /tmp/pwn)
OSC 8 wire form:              \x1b]8;;man:ls(-P /usr/bin/touch /tmp/pwn)\x1b\\click here\x1b]8;;\x1b\\
```

mat-mo:
The less-osc8-open.sh script was vulnerable to argument injection when handling OSC 8 man: URIs. This change adds validation to the section identifier to ensure it cannot be parsed as a command option by man.

### DIFF
--- a/less-osc8-open.sh
+++ b/less-osc8-open.sh
@@ -43,6 +43,15 @@ open_man() {
 	man:?*\(*\) )
 		sect=${1#*\(}; sect=${sect%?}
 		name=${1#man:}; name=${name%%\(*}
+
+                case $sect in
+                    ''|[1-9]|[1-9][A-Za-z0-9]*)
+                        ;;
+                    *)
+                        echo "invalid man link section: $sect" >&2
+                        exit 1
+                        ;;
+                esac
 		man ${sect:+"$sect"} "$name"
 		;;
 
--- a/less-osc8-open.sh
+++ b/less-osc8-open.sh
@@ -46,13 +46,16 @@ open_man() {
 
                 case $sect in
                     ''|[1-9]|[1-9][A-Za-z0-9]*)
+                       case $sect in
+                           -* ) echo "invalid man link section: $sect" >&2; exit 1 ;;
+                        esac
                         ;;
                     *)
                         echo "invalid man link section: $sect" >&2
                         exit 1
                         ;;
                 esac
-		man ${sect:+"$sect"} "$name"
+		man ${sect:+"$sect"} -- "$name"
 		;;
 
 	man:?*)


avih:
As noted at #779, the most robust solution would be simply this change:
```diff
- 		man ${sect:+"$sect"} "$name"
+ 		man -- ${sect:+"$sect"} "$name"
```

Anything else gets us into territories where we can't be 100% sure we're doing the right thing, IMO.

The problem with `--` is that we also can't know that every version of `man` supports it. POSIX says it should, but a form which we support (man SECT NAME) is already not POSIX, so maybe we can't actually know that every version of man supports `--`.

Since that issue was reported at #779 and that issue includes much more info (reasoning, factors) than here, I think we should continue to discuss it there, figure out the best approach, and then this (or another) PR can address it.

avih:
Also, note that since the script doesn't (try to) control the locale, then `[1-9]|[1-9][A-Za-z0-9]` is not guaranteed to be what you expect it to be, i.e. the values which this set covers can differ between locales.

Preferably, we should use my suggestion without anything else, assuming everyone agrees that it's indeed the most robust approach.

mat-mo:
I agree, my only concern is indeed the POSIX angle. I still think we should go your route, and clarify it only works when man supports --.

I don't know what happens on MS-DOS for example.

PR #788, opened 2026-05-27
Fix typo in uninstall rule.

mmuetzel:
The `uninstall` Makefile rule attempts to remove the binaries in the libexec folder using the wrong variables.
Fix that by using the same variables that are used in the `install` Makefile rule for these binaries.


mmuetzel:
Thank you for merging. 👍

Issue #778, opened 2026-05-25
Adding SPDX-License-Identifier headers to source files


illuusio:
It would help a lot to have SPDX-License-Identifiers in source files headers. Those would be great for maintaining SBOM information.

If it's okay to add them I can do it

gwsw:
Go ahead and make a PR, based in the post700 branch.

PR #792, opened 2026-06-01
Add SPDX-License-Identifier tags to less source files

illuusio:
Less is mostly licensed under
 * GPL-3.0-or-later OR BSD-2-Clause

But there is couple exceptions:
 * **regexp.c is** under: Spencer-86
 * **scrsize.c** is under: X11 AND (GPL-3.0-or-later OR BSD-2-Clause)

This fixes: #778

### DIFF
--- a/brac.c
+++ b/brac.c
@@ -1,4 +1,6 @@
 /*
+ * SPDX-License-Identifier: GPL-3.0-or-later OR BSD-2-Clause
+ *
  * Copyright (C) 1984-2026  Mark Nudelman
  *
  * You may distribute under the terms of either the GNU General Public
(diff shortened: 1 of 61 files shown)


SpecLad:
Note that the Less license is not the same as the 2-Clause BSD License. Where BSD-2-Clause says:

> Redistributions in binary form must reproduce the above copyright notice, this list of conditions and the following disclaimer in the documentation and/or other materials provided with the distribution.

the Less license says:

> Redistributions in binary form must reproduce the above copyright notice in the documentation and/or other materials provided with the distribution.

illuusio:
Thank you pointing this out and sorry for my unaccurancy! I trusted too much Scancode that said it's BSD-2-Clause but after reading mainly all variants which it contains it true that Less has unique  variant of BSD-2-Clause.  I think I'll try to commit this license to Scancode and then try to get it into SPDX (long way). It's bit like like there is BSD-3-Clause-flex. I'll update this to `LicenseRef-Less` or something we'll work out.

Issue #719, opened 2026-01-29
Infinite loop at 100% CPU when tty unavailable and binary file prompts for input


LangLangBart:
### Steps to Reproduce

1. Run the command below:

```bash
{ less /bin/sh & echo $! > /tmp/p; sleep 1; } 2>&1 | cat >/dev/null &
```

2. Check CPU usage for less:

```bash
ps -o pid,ppid,pcpu,state,command -p $(cat /tmp/p)
```

3. Clean up:

```bash
kill -9 $(cat /tmp/p)
```

### Observed Behavior

Never-ending less process with high CPU usage:

```txt
  PID  PPID  %CPU STAT COMMAND
96965     1 100.0 R    less /bin/sh
```

what seems to be happening:

1. `less /bin/sh &` runs with stdout piped → `is_tty=0`
2. Since `is_tty=0`, `main.c` takes the "just cat to output" path
3. This path does **NOT** call `open_getchr()`
4. So `tty` stays at its default value of 0 (stdin)
5. BUT `/bin/sh` is a binary file, so `edit.c` calls `query()` asking "See it anyway?"
6. `query()` calls `getchr()` to get user response
7. `getchr()` calls `iread(tty=0, ...)` - reading from stdin
8. stdin is closed/EOF in the background scenario
9. `read()` returns 0 forever → infinite busy loop at 100% CPU

### Expected Behavior

`less` should not become an endlessly running high CPU usage process. It should either:

- Quit with an error when tty input is unavailable
- Not prompt for binary file confirmation when output is not a tty

### Regression Information

The `consecutive_nulls` kludge that previously caught this case was disabled

- Bisected to commit: 8db7960040647b94812309d4b48ea9614bd04c55

### Additional Information

This issue was discovered after noticing a `less` process heavily hitting the
CPU when using fzf's preview command.

- Ref: https://github.com/junegunn/fzf/issues/4667

gwsw:
I agree that there's no need to attempt to ask the user for confirmation when the output is not a tty.
Fixed in 0f41a20b9a34f84d3f1751147511f036f5db6502.

LangLangBart:
thanks for the quick fix

I would prefer it to display a binary file message instead, e.g. `"/bin/sh" may be a binary file (use -f to open)`

https://github.com/gwsw/less/blob/0a8419135fd7c86061f154af14d903e54a646ae9/less.nro.VER#L1073-L1078


[...]

Issue #558, opened 2024-08-09
Regression - less does not read whole pipe when stdin is /dev/null


mhlavink:
Hi,

we found and issue that is related to commit 953257f9030d29dd930ed3e76fd5811681b672b1 that was identified using bisect. 

less does not read whole piped input when stdin is /dev/null

We don't update all packages that often, so we've noticed this issue just now, when I've updated less and tests started to fail. I've tried to search for this issue report and found some similar (mostly old) issues, but this seems different.

Reproducer:
1) prepare zip file
```
$ date >file
$ zip file.zip file
```

2) run
`$ less file.zip </dev/null | cat`

Actual result: 
only partial output, one line mostly when run plain, two or two and half when run with strace:
```
Archive:  file.zip
```

Expected result:
```
Archive:  file.zip
Zip file size: 190 bytes, number of entries: 1
-rw-r--r--  3.0 unx       32 tx stor 24-Aug-09 05:25 file
1 file, 32 bytes uncompressed, 32 bytes compressed:  0.0%
``` 

tail of strace output used latest git:
```
readlink("/home/tester/less/file.zip", 0x7fff4e3771e0, 1023) = -1 EINVAL (Invalid argument)
pipe2([3<pipe:[22383]>, 4<pipe:[22383]>], O_CLOEXEC) = 0
prlimit64(0, RLIMIT_NOFILE, NULL, {rlim_cur=1024, rlim_max=512*1024}) = 0
prlimit64(0, RLIMIT_NOFILE, NULL, {rlim_cur=1024, rlim_max=512*1024}) = 0
mmap(NULL, 36864, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS|MAP_STACK, -1, 0) = 0x7f631ebad000
rt_sigprocmask(SIG_BLOCK, ~[], [], 8)   = 0
clone3({flags=CLONE_VM|CLONE_VFORK|CLONE_CLEAR_SIGHAND, exit_signal=SIGCHLD, stack=0x7f631ebad000, stack_size=0x9000}, 88) = 19582
munmap(0x7f631ebad000, 36864)           = 0
rt_sigprocmask(SIG_SETMASK, [], NULL, 8) = 0
close(4<pipe:[22383]>)                  = 0
fcntl(3<pipe:[22383]>, F_SETFD, 0)      = 0
read(3<pipe:[22383]>, "A", 1)           = 1
[...]

PR #794, opened 2026-06-05
Fix tty EOF handling

### LINKED ISSUE #793: less spins at 100% CPU when tty read returns EOF after pty close
### Summary

`less` can spin at 100% CPU when keyboard input is read from a tty whose underlying pty has been closed.

In `ttyin.c:getchr()`, the `LESS_IREAD_TTY` path repeatedly calls `iread(tty, &uc, sizeof(char))` until it gets one byte. It handles `READ_INTR` and negative results, but it does not handle `read(2)` returning `0` for EOF. `os.c:iread()` returns that EOF value unchanged, so `getchr()` keeps looping while `result != 1` even though no byte can arrive.

This is visible in pty-based callers when the pty master is closed while `less` is waiting for keyboard input. It is especially painful when `LESSOPEN`/lesspipe-style input preprocessors are in use, because `less` not exiting can also delay child cleanup.

### Steps to reproduce

One way to observe the issue is from a pty-owning caller that starts `less`, then closes the pty master while `less` is waiting for input. This has shown up in terminal UI preview/task integrations that use a pty, such as fzf/lazygit-style preview processes.

At the code level, the relevant path is:

1. `ttyin.c:getchr()` calls `iread(tty, &uc, sizeof(char))`.
2. The pty closes and the underlying `read(2)` returns `0`.
3. `os.c:iread()` returns `0` unchanged.
4. `getchr()` does not handle `result == 0`, so the `while (result != 1)` loop immediately retries.

### Expected behavior

If reading keyboard input from the tty returns EOF, `less` should stop waiting for a byte and exit/clean up instead of looping indefinitely.

### Actual behavior

`getchr()` handles `READ_INTR` and negative errors, but `result == 0` falls through and repeats the loop. This can leave `less` consuming CPU until killed externally.

### Environment

Observed in current upstream source:

- `ttyin.c:getchr()` loops while `result != 1` and has no `result == 0` case.
- `os.c:iread()` is a thin wrapper around `read(2)` and returns `0` unchanged.

### Related issues

This appears related to, but distinct from, prior EOF/busy-loop issues:

- #558 handled an EOF busy loop in the polling path.
- #719 handled a binary-file prompt path when output is not a tty.
- junegunn/fzf#4667 reported preview processes involving `less` consuming CPU after preview shutdown.

Those fixes do not appear to handle the `ttyin.c:getchr()` `result == 0` path.

HaleTom:
### Issue for this PR

Closes #793.

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

`getchr()` reads keyboard input in a loop until `iread()` returns one
byte. It already handles interrupted reads and negative errors, but it
does not handle EOF. Since `iread()` returns `read(2)` EOF unchanged,
closing the underlying tty/pty can leave `less` looping forever while no
input byte can arrive.

This change treats EOF from the tty input read as a fatal input condition
and exits through `quit(QUIT_ERROR)`, matching the nearby negative-error
path and avoiding a busy loop.

I chose direct `quit()` rather than returning `READ_INTR` because
`getccu()` converts negative `getchr()` results to NUL and loops while no
signal bit is set. Returning `READ_INTR` for EOF without also setting a
signal bit can move the spin to the caller instead of stopping it.

Related prior fixes/discussion:

- #558 fixed a separate EOF busy loop in the polling path.
- #719 fixed a separate prompt path when output is not a tty.
- junegunn/fzf#4667 reported preview processes involving `less` consuming
  CPU after preview shutdown.

### How did you verify your code works?

Built from a fresh checkout using the documented development path:

```sh
make -f Makefile.aut distfiles
sh configure
make
./less --version
```

The build completed and the built `less` binary reported version 704.

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

PR template not found in `.github/pull_request_template.md`,
`.github/PULL_REQUEST_TEMPLATE/`, `.github/`, or `CONTRIBUTING.md`.

### DIFF
--- a/ttyin.c
+++ b/ttyin.c
@@ -216,14 +216,6 @@ public int getchr(void)
 		}
 		if (result == READ_INTR)
 			return (READ_INTR);
-		if (result == 0)
-		{
-			/*
-			 * EOF on the tty means there is no more keyboard input.
-			 * Don't loop forever waiting for a byte which cannot arrive.
-			 */
-			quit(QUIT_ERROR);
-		}
 		if (result < 0)
 		{
 			/*
--- a/ttyin.c
+++ b/ttyin.c
@@ -216,6 +216,14 @@ public int getchr(void)
 		}
 		if (result == READ_INTR)
 			return (READ_INTR);
+		if (result == 0)
+		{
+			/*
+			 * EOF on the tty means there is no more keyboard input.
+			 * Don't loop forever waiting for a byte which cannot arrive.
+			 */
+			quit(QUIT_ERROR);
+		}
 		if (result < 0)
 		{
 			/*


avih:
> Since `iread()` returns `read(2)` EOF unchanged,
> closing the underlying tty/pty can leave `less` looping forever while no
> input byte can arrive.

I'm assuming you're right, but I'm not seeing anything which would allow me to confirm it myself, other than trying to figure out myself at which scenarios this issue might actually manifest in practice.

It would be a lot nicer if there was also a concrete example of how to reproduce such issue, probably also at the commit message (because whoever clones the repo and looks at the git history doesn't see the discussions on github).

The best thing would probably be to first file a bug report with steps to reproduce, and once the bug is confirmed by others and they can reproduce it, then it's time for someone to fix it.

But since you already opened a PR, then the commit message should probably include steps to reproduce the issue which others could follow with ease.

HaleTom:
@avih commit message in 1f92725af3ac23a7bcf45ff5a1490069bf836570 now includes full code for reproduction

PR #797, opened 2026-06-14
Exit with a failure status in more cases

SpecLad:
The following conditions are fatal, yet less currently handles them by returning a success status. Return a failure status instead.

* Missing input filename when stdin is a terminal.

* Missing option argument (e.g. `less -b`).

* Not being able to open any input files in non-interactive mode (note that in interactive mode, this already returns a failure status).

### DIFF
--- a/edit.c
+++ b/edit.c
@@ -919,7 +919,7 @@ public int edit_stdin(void)
 	if (isatty(fd0))
 	{
 		error("Missing filename (\"less --help\" for help)", NULL_PARG);
-		quit(QUIT_OK);
+		quit(QUIT_ERROR);
 	}
 	return (edit("-"));
 }
--- a/main.c
+++ b/main.c
@@ -302,7 +302,7 @@ int main(int argc, constant char *argv[])
 		 * following string, but there was no following string.
 		 */
 		nopendopt();
-		quit(QUIT_OK);
+		quit(QUIT_ERROR);
 	}
 
 	if (is_tty)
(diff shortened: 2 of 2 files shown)


avih:
> remove a confusing `exit(0)` in `lessecho.c`. It's unreachable

This is quite likely to please some compilers which get grumpy when a function which must a value - doesn't return a value (because they don't do the analysis which you just did to understand that it's actually a no-return function).

A better solution IMHO would be to add a comment `/* unreachable */`.

EDIT: actually, I don't see at the changes this `exit(0)` which you claim to have removed. Did you forget to actually remove it?

SpecLad:
> This is quite likely to please some compilers which get grumpy when a function which must a value - doesn't return a value (because they don't do the analysis which you just did to understand that it's actually a no-return function).

In this case `main` does still return a value at the end.

> EDIT: actually, I don't see at the changes this `exit(0)` which you claim to have removed. Did you forget to actually remove it?

Oh, I made a mistake in the description - it's actually a `return` rather than an `exit` call. Fixed now.

SpecLad:
Ah, never mind, the warning this triggers is `-Wimplicit-fallthrough`. I'm just going to leave it alone then.

avih:
> Ah, never mind, the warning this triggers is `-Wimplicit-fallthrough`. I'm just going to leave it alone then.

Exactly what I was about to write that compilers might do. But you could still add an "unreachable" comment like I suggested initially, to make it clear to the next person which looks at the code, because it already confused you once, so it could still confuse others.

SpecLad:
I think I'd rather keep this PR to its main scope. I clearly shouldn't have touched that `return` in the first place. :slightly_smiling_face:

PR #798, opened 2026-06-14
Add a GitHub Actions workflow that tests less on Linux

SpecLad:
I don't know if you'd be interested in something like this, but I thought it could be useful for testing incoming pull requests and commits.

If this is accepted, I'll try to send a followup PR to enable testing on Windows as well.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,61 @@
+name: Main
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+
+permissions:
+  contents: read
+
+jobs:
+  dist:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v7
+      with:
+        persist-credentials: false
+
+    - name: Build distribution
+      id: build-dist
+      # Since we don't have a PGP key here, replace gpg with a no-op.
+      run: |
+        make -f Makefile.aut dist GPG=:
+        echo "basename=$(basename release/less-*)" > "$GITHUB_OUTPUT"
+
+    - uses: actions/upload-artifact@v7
+      with:
+        archive: false
+        path: release/${{ steps.build-dist.outputs.basename }}/${{ steps.build-dist.outputs.basename }}-beta.tar.gz
+
+    outputs:
+      basename: ${{ steps.build-dist.outputs.basename }}
+
+  linux:
+    runs-on: ubuntu-latest
+
+    needs: dist
+
+    steps:
+    - uses: actions/download-artifact@v8
+      with:
+        name: ${{ needs.dist.outputs.basename }}-beta.tar.gz
+
+    - name: Unpack
+      env:
+        BASENAME: ${{ needs.dist.outputs.basename }}
+      run: tar -xzf "$BASENAME-beta.tar.gz"
+
+    - name: Configure
+      working-directory: ${{ needs.dist.outputs.basename }}
+      run: ./configure
+
+    - name: Make
+      working-directory: ${{ needs.dist.outputs.basename }}
+      run: make LESSTEST=1
+
+    - name: Test
+      working-directory: ${{ needs.dist.outputs.basename }}
+      run: make check LESSTEST=1
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -24,11 +24,12 @@ jobs:
       run: |
         make -f Makefile.aut dist GPG=:
         echo "basename=$(basename release/less-*)" > "$GITHUB_OUTPUT"
+        mv -v release/less-*/*.tar.gz less-githubtest.tar.gz
 
     - uses: actions/upload-artifact@v7
       with:
         archive: false
-        path: release/${{ steps.build-dist.outputs.basename }}/${{ steps.build-dist.outputs.basename }}-beta.tar.gz
+        path: less-githubtest.tar.gz
 
     outputs:
       basename: ${{ steps.build-dist.outputs.basename }}
@@ -41,12 +42,10 @@ jobs:
     steps:
     - uses: actions/download-artifact@v8
       with:
-        name: ${{ needs.dist.outputs.basename }}-beta.tar.gz
+        name: less-githubtest.tar.gz
 
     - name: Unpack
-      env:
-        BASENAME: ${{ needs.dist.outputs.basename }}
-      run: tar -xzf "$BASENAME-beta.tar.gz"
+      run: tar -xzf less-githubtest.tar.gz
 
     - name: Configure
       working-directory: ${{ needs.dist.outputs.basename }}
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -66,4 +66,4 @@ jobs:
 
     - name: Test
       working-directory: ${{ needs.dist.outputs.basename }}
-      run: make check LESSTEST=1
+      run: make check
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -54,6 +54,14 @@ jobs:
 
     - name: Make
       working-directory: ${{ needs.dist.outputs.basename }}
+      run: make
+
+    - name: Run sanity check
+      working-directory: ${{ needs.dist.outputs.basename }}
+      run: ./less README > README.out && cmp README README.out
+
+    - name: Make (LESSTEST)
+      working-directory: ${{ needs.dist.outputs.basename }}
       run: make LESSTEST=1
 
     - name: Test
--- a/Makefile.aut
+++ b/Makefile.aut
@@ -163,8 +163,7 @@ dist: ${DISTFILES}
 	echo "Creating $$RELDIR/$$TARF" && \
 	tar -cf - $$LESSREL | gzip -c >$$RELDIR/$$TARF && \
 	echo "Signing $$RELDIR/$$TARF" && \
-	${GPG} --detach-sign $$RELDIR/$$TARF && \
-	mv $$RELDIR/$$TARF.sig $$RELDIR/$$LESSREL.sig && \
+	${GPG} --detach-sign --output=$$RELDIR/$$LESSREL.sig $$RELDIR/$$TARF && \
 	echo "Creating $$RELDIR/$$ZIPF" && \
 	zip -rq $$RELDIR/$$ZIPF $$LESSREL && \
 	rm -rf $$LESSREL


SpecLad:
You can see what the workflow runs will look like in my test PR: SpecLad/less#1.

gwsw:
I'm not very familiar with github workflows, so I'm not sure I completely understand what this does. I have a few questions.
1. Where does the package get uploaded to by upload-artifact? Is this uploaded package visible to other users?
2. Is the uploaded package named as usual for less packages? In particular, is the version number (which seems to come from ```needs.dist.outputs.basename```) taken from the version.c file? I'm concerned that if multiple pushes are done between updates to that version number, there will be multiple packages with the same name. This could cause confusion, especially if these packages are visible to other users.
3. If a PR is submitted and the test is run successfully, and then later a commit changes some code in a way that invalidates the PR, is that detected somehow? Is a test run against all open PRs whenever a commit is done?
4. It looks like the test runs make with LESSTEST=1, but never runs a normal make without LESSTEST set. It seems like it would be useful to run a normal make as well as a LESSTEST make.
5. Minor point: it's not necessary to explicitly set LESSTEST=1 when running "make check". LESSTEST is set internally in the check action.

avih:
The archive name should probably include at the very least also the git hash, and preferably also the build date, to make it easier to figure out in a glance.

SpecLad:
> Where does the package get uploaded to by upload-artifact?

It gets attached to the workflow run. If you open my test PR, you can find it if you click on the green checkmark next to the commit, click "Details" next to one of the jobs, and then go to Summary.

Artifacts like this are retained for 90 days.

> Is this uploaded package visible to other users?

Yes, it's visible to everybody.

> Is the uploaded package named as usual for less packages? In particular, is the version number (which seems to come from needs.dist.outputs.basename) taken from the version.c file?

I didn't change anything about the package naming, so they just get named what they would normally (e.g. `less-705-beta.tar.gz`).

If you think this could be confusing, then I can rename the file to something that is obviously not a release (e.g. `less-dev.tar.gz`).

> If a PR is submitted and the test is run successfully, and then later a commit changes some code in a way that invalidates the PR, is that detected somehow? Is a test run against all open PRs whenever a commit is done?

No, a PR is only retested if it is updated (i.e. a new commit is pushed into the source branch). If you want to make sure each PR is tested with the latest `master` before it's merge, I think you can do it with [merge queues](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue), but I've never used this feature myself, so I can't help you there.

> It looks like the test runs make with LESSTEST=1, but never runs a normal make without LESSTEST set. It seems like it would be useful to run a normal make as well as a LESSTEST make.

Sure, I can do that.

> Minor point: it's not necessary to explicitly set LESSTEST=1 when running "make check". LESSTEST is set internally in the check action.

Okay, I'll update the PR and remove it.

SpecLad:
Alright, I added a build without lesstest (and a small sanity check), and removed the redundant `LESSTEST=1` option. Since you didn't say anything about the package name, I've refrained from making changes there for now; let me know what you think.

gwsw:
I think this looks reasonable, although I would prefer that the package name be different from any released package name. How about ```less-githubtest.tar.gz```?

avih:
I think it also still doesn't include the git hash and build date.

These are important. Otherwise one would download some builds to examine locally, and it quickly gets out of hand if it's not obvious from which commit it was built.

gwsw:
I think the git hash appears in the log of the workflow, doesn't it? Maybe that is sufficient, but I don't object to having the hash in the file name as well. However, I would not expect anyone to be downloading the tarball generated by this workflow. Anyone working on the less source would already have their own clone of the repository and would be working from there.

avih:
If you don't expect anyone to download it, then disable the upload part of the PR.

But history (elsewhere) shows that sometimes these need to be downloaded and examined, and this PR does upload the binary to a place where it can be downloaded, so it's much better that the file name makes it easy to identify the content.

Besides the hash and date, typicallly the CI build file name also includes the platform and other attributes (xBSD, linux foo, linux bar, 32 bit windows, 64 bit wibdows, with PCRE, without, etc).

It's just useful that the file is identifiable, and if you don't think anyone will download it then don't upload it.

gwsw:
I know little about how workflows work, but I thought that it was being uploaded so that the build step could access it, by calling download-artifact. If there's another way for the build step to access the package without uploading it, I think I would prefer that. I don't like there to be ways for users to download what appears to be a less release but isn't really one.

avih:
I'm also not very familiar with CI setup, but there are plenty of setups which do build and run tests but don't make anything available for download.

I'm not fond of those because, as noted, sometimes it can be very useful to download and examine the build.

I don't share your concern that it looks like a release build. It's a CI build for testing (the build process and sometimes the build binary), which is a known concept, and non-release build already include a large warning anyway.

It can be useful for testing if someone is not able to build it on their own, for instance. This is very common elsewhere where users can download and test a build with new features or fixes and report back. It's useful to be able to download it.

SpecLad:
@gwsw 

> How about `less-githubtest.tar.gz`?

Okay, I changed the name to this.

@avih FYI, the file that gets uploaded is not a _build_ really, it's only a source distribution. I'm not expecting anyone to download these files, let alone download so many of them that they lose track of where they came from.

The reason it gets uploaded as an artifact is so that it can be transferred between jobs. And the reason there are two separate build jobs is so that the build is run in a clean environment that doesn't include the original source repository. This helps catch mistakes where a source file is not included in the distribution by mistake, and in fact, it just did (see the last commit)!

avih:
@SpecLad Could you please clarify why the recent changes were made? are they addressing some review comment? (which one/ones?) are they fixing some unrelated issue?

For instance, https://github.com/gwsw/less/pull/798/commits/ec5fb097444a68befcb2fee55721cfc3464589de adds `lmsg.h` to `Makefile.aut`, presumably due to this CI failure: https://github.com/SpecLad/less/actions/runs/28962534515/job/85937427862#step:5:8

But I don't think any of your other commits touch this file, so why is it part of this PR? (not saying it shouldn't be here, just that I don't yet understand why).

SpecLad:
@avih I mean... seems like you've answered your own question there. I added `lmsg.h` to the list of distribution files because the build doesn't work without it.

avih:
> I added `lmsg.h` to the list of distribution files because the build doesn't work without it.

I see that, but does that fix a problem in one of your commits? or does that try to fix an unrelated issue in "less"?

Presumably the distribution is not broken in master, so why was it broken with this PR?

SpecLad:
> I see that, but does that fix a problem in one of your commits? or does that try to fix an unrelated issue in "less"?

The latter.

> Presumably the distribution is not broken in master, so why was it broken with this PR?

It is, in fact, broken in master.

avih:
> It is, in fact, broken in master.

Is that a recent thing? I'm not seeing any recent changes in master which might have broken it.

gwsw:
The problem was created in 9b3eaff77b05244f9b69ed4039554870221530a9 when I added lmsg.c and lmsg.h. I added lmsg.c to the list of distribution files in Makefile.aut but neglected to add lmsg.h.

avih:
> The problem was created in [9b3eaff](https://github.com/gwsw/less/commit/9b3eaff77b05244f9b69ed4039554870221530a9) when I added lmsg.c and lmsg.h

Thanks.

In such case, the fix should probably not be hidden at an unrelated PR without explaining it like you did just now.

Otherwise it looks like it's part of the changes which the PR does in order to support the CI build.

gwsw:
I agree with this. I have done the lmsg.h fix in a651431d9551428b56f3eda83c6faba3744ba6d2, so it can be removed from this PR.

avih:
> The problem was created in [9b3eaff](https://github.com/gwsw/less/commit/9b3eaff77b05244f9b69ed4039554870221530a9) when I added lmsg.c and lmsg.h

Hmm.. off topic for this PR, but I looked at that commit now (apparently I missed it, and it also includes the sed tab thing, I didn't realize it's a new thing when the omnios build failed). It's a pretty big commit. Quite a few new dependency files too at the makefiles (headers). Didn't look at it too closely yet, but would it be reasonable to try and reduce the number of individual headers of this "system"?

EDIT: also, it adds generated files. Which means that the mingw/msvc build don't yet generate them. They probably should be able to, because changing messages text is part of development, and until now the mingw/msvc makefiles didn't depend on `Makefile.aut`, but now they do after message changes.

SpecLad:
> I agree with this. I have done the lmsg.h fix in [a651431](https://github.com/gwsw/less/commit/a651431d9551428b56f3eda83c6faba3744ba6d2), so it can be removed from this PR.

And removed.

gwsw:
>would it be reasonable to try and reduce the number of individual headers of this "system"?

I'm not sure I understand. I think only one new header file was added, lmsg.h. Are you referring to something else?

The contents of lmsg.h could probably have been added directly into less.h. Not sure if it's worth it at this point. The change is large in number of lines, but almost all of the changes are just replacing a string constant with a ```LM(xxx)``` expression.

avih:
> I'm not sure I understand. I think only one new header file was added, lmsg.h. Are you referring to something else?

I'm seeing 5 additional dependencis: `lang.h` `lmsg.h` `lessmsg.inc` `pattern.h` `xbuf.h`. Like this change in `Makefile.wng`, which also exists in other makefiles in that commit:
```diff
-${OBJ}: less.h defines.h funcs.h
+$(OBJ): less.h funcs.h defines.h lang.h lmsg.h lessmsg.inc pattern.h xbuf.h
```

Or maybe some of these changes unrelated to the messages system (pattern.h, xbuf.h, lang.h), and the only ones which are actually related are lmsg.h and lessmsg.inc ?

Also, correct me if I'm wrong, but I'm not seeing `lessmsg.inc` at the sources, so that means that the mingw/msvc makefiles can't build a git snapshot from git sources on their own, unless `Makefile.aut` is used to generate it? (I _think_ it's just a one liner at Makefile.aut, but it's still only at that file)

(sorry, I know this is off topic, but I just became aware of this new system following your earlier comment that the distribution lmsg.h fix is indeed a recent breakage)

gwsw:
>I'm seeing 5 additional dependencis

When I went into the various makefiles to add the lmsg.h dependency, I noticed tht some long existent dependencies were missing from some of the makefiles, so I brought them all up to date. If you compare the Makefiles just before 9b3eaff77b05244f9b69ed4039554870221530a9, you'll see that the OBJ dependencies are inconsistent.

>the only ones which are actually related are lmsg.h and lessmsg.inc ?

Yes.

>mingw/msvc makefiles can't build a git snapshot from git sources on their own, unless Makefile.aut is used to generate it? 

Yes, that's currently the case. Maybe buildgen.c could be enhanced to generate lessmsg.inc but I haven't done that yet.

avih:
> Yes, that's currently the case.

Thanks (I just didn't have time to follow less master, or else I'd have absolutely noticed that breakage earlier, and I also didn't look at it closely, so I just noticed many new deps at the makefiles...).

> Maybe buildgen.c could be enhanced to generate lessmsg.inc but I haven't done that yet.

We should do that, yeah, and probably also the generator from Makefile.aut copied to Makefile.wng - which already depends on posix userland, so it should be OK.

SpecLad:
@gwsw So are we good here? I think I've addressed all of your comments.

PR #804, opened 2026-07-10
Add a Windows job to the GitHub workflow

SpecLad:
There's no support for building lesstest on Windows, so we can't run the tests. Just run a basic sanity check instead.

To improve coverage, use the ZIP archive instead of the tar archive like the Linux job.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -83,29 +83,37 @@ jobs:
     strategy:
       matrix:
         toolchain: [mingw-w64, msvc]
+        source: [dist, git]
+
+    defaults:
+      run:
+        working-directory: ${{ case(matrix.source == 'dist', needs.dist.outputs.basename, '.') }}
 
     steps:
     - uses: actions/download-artifact@v8
+      if: matrix.source == 'dist'
       with:
         # Since the artifact is a ZIP file, it will automatically be unpacked.
         name: less-githubtest.zip
 
+    - uses: actions/checkout@v7
+      if: matrix.source == 'git'
+
     - name: Make with Mingw-w64
       if: matrix.toolchain == 'mingw-w64'
-      working-directory: ${{ needs.dist.outputs.basename }}
       run: |
         mingw32-make -j4 -f Makefile.wng
 
     - name: Make with Visual C++
       if: matrix.toolchain == 'msvc'
-      working-directory: ${{ needs.dist.outputs.basename }}
+      env:
+        TARGETS: ${{ case(matrix.source == 'dist', 'all', 'generated all') }}
       run: |
         $vsPath = & "${Env:ProgramFiles(x86)}\Microsoft Visual Studio\Installer\vswhere.exe" -latest -property installationPath
         & "$vsPath\Common7\Tools\Launch-VsDevShell.ps1" -Arch amd64
-        nmake -f Makefile.wnm
+        nmake -f Makefile.wnm (-split $Env:TARGETS)
 
     - name: Run sanity check
-      working-directory: ${{ needs.dist.outputs.basename }}
       run: |
         .\less.exe README > README.out
         cmd /c fc /b README README.out
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -25,11 +25,19 @@ jobs:
         make -f Makefile.aut dist GPG=:
         echo "basename=$(basename release/less-*)" > "$GITHUB_OUTPUT"
         mv -v release/less-*/*.tar.gz less-githubtest.tar.gz
+        mv -v release/less-*/*.zip less-githubtest.zip
 
-    - uses: actions/upload-artifact@v7
+    - name: Upload tarball
+      uses: actions/upload-artifact@v7
       with:
         archive: false
         path: less-githubtest.tar.gz
+
+    - name: Upload ZIP file
+      uses: actions/upload-artifact@v7
+      with:
+        archive: false
+        path: less-githubtest.zip
 
     outputs:
       basename: ${{ steps.build-dist.outputs.basename }}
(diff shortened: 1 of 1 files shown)


avih:
Thanks.

Note that this tests only a release build, but both Makefile.wnm and Makefile.wng support also building a git snapshot - which doesn't include generated files (currently funcs.h, help.c, lessmsg.inc).

With Makefile.wng a normal `make -f Makefile.wng` auto-generates these files if missing, and with Makefile.wnm it needs an additional preliminary run of `nmake -f Makefile.wnm generated` (I might make that automatic too - like the mingw file, not sure yet).

So it's worth testing too, because it can get broken, like it was recently with `lessmsg.inc`.

SpecLad:
@avih Okay, fair point. I implemented it.

avih:
Looks good, thanks.

Unrelated, I still think that making the resulting artifacts available for download (until it expires on its own) can be useful, both for devs to examine it (and test builds which they can't build themselves, like different compiler version etc) and for user who can't build it on their own to test recent changes and fixes.

Being able to get early feedback from end users - rather than only at the next release, is a very big advantage.

avih:
Interesting. mingw-git failed: https://github.com/gwsw/less/actions/runs/29205050032/job/86682934143
```
decode.c: In function 'lesskey_text':
decode.c:1421:23: error: implicit declaration of function 'parse_lesskey_content'; did you mean 'lesskey_content'? [-Wimplicit-function-declaration]
 1421 |         r = content ? parse_lesskey_content(filename, &tables) : parse_lesskey(filename, &tables);
      |                       ^~~~~~~~~~~~~~~~~~~~~
      |                       lesskey_content
decode.c:1421:66: error: implicit declaration of function 'parse_lesskey'; did you mean 'new_lesskey'? [-Wimplicit-function-declaration]
 1421 |         r = content ? parse_lesskey_content(filename, &tables) : parse_lesskey(filename, &tables);
      |                                                                  ^~~~~~~~~~~~~
      |                                                                  new_lesskey
mingw32-make: *** [Makefile.wng:110: decode.o] Error 1
mingw32-make: *** Waiting for unfinished jobs....
Error: Process completed with exit code 1.
```

I wonder if that's a makefile rules dependency issue which got broken by the parallel `-j4`.

avih:
> I wonder if that's a makefile rules dependency issue which got broken by the parallel `-j4`.

Nope. It's an actual breakage on master.

Goot catch on first run on master!

avih:
It's a regression of 4c010b1a85d25244eb0ee1c76008fd8fb45449aa where `lesskey_parse.c` was added to `Makefile.aut` as a source file but not added to any of the other makefies.

gwsw:
lesskey_parse.o has always been in all of the Makefiles. But the ones that build funcs.h themselves like Makefile.wng also need it listed as a source file. Makefile.wnm avoids this problem by using "*.c" as the input to buildgen, which probably isn't ideal since that includes files that aren't part of less, like buildgen.c, lessecho.c, etc as well as any other .c files you happen to have lying around. However most such files don't use the "public" keyword, so usually extra declarations don't get added to funcs.h.

avih:
Right. I think it would be better to just have list of source files and do the automatic `.o` files which `make` can do to have just one list which can't go out of sync with the other? kinda the oposit of what `Makefile.wnm` does (which only specifies the `.obj` files).

gwsw:
Yeah, the lesstest makefile does this with the line
```
LESSTEST_OBJ = $(patsubst %.c,%.o,$(LESSTEST_SRC))
```
I don't know how standard this is. I haven't tried it with Microsoft's nmake, so I don't know if it works there.

avih:
The question about the mingw makefile because it has two lists which can get out of sync - as they just did. The msvc makefile already has one list, albeit only of obj files and not of source files.

gwsw:
Well, Makefile.wng could be changed to just use ```*.c``` for the source list like Makefile.wnm does, but I don't think that's a good idea. I think both Makefile.wng and Makefile.wnm should be changed to use patsubst to derive one list from the other, assuming that patsubst works in the relevant make programs.

avih:
Yes, both having only one list of source files would be ideal.

But the msvc one already has one list, and we don't know whether it can be changed from obj files to src files.

But the mingw has two lists and can probably use the patsubst thing because I think it already depends on gnu make features anyway (the `ifeq` things).

avih:
Also, I wasn't referring to patsubst specifically. That's one way to create one list from another list in gnu make, but I think there's also a posix way to specify only source files and then compile them to `.o` files. I think something with `%` which I don't remember exactly.

gwsw:
FWIW, ChatGPT tells me that there's no way to do it in POSIX make. It suggests patsubst, or the equivalent
```
OBJ = $(SRC:.c=.o)
```
but says that this only works in gnu and BSD and some other makes, but is not specified by POSIX.

avih:
Whatever you prefer. I think I'd use the patsubst thing because it's already used elsewhere in less makefiles, and Makefile.wng already depends on gnu make as far as I can tell.

But if you think `OBJ = $(SRC:.c=.o)` is more portable then sure.

avih:
Well, I tried the patsubst things, which works, but I found out two things:
1. The list of sources accidentally includes `scrsize.c`, which fails to compile because it can't compile on windows (apparently an X11 thing).
2. the obj files need the patsubst of the src files plus `help.o` (which was at the obj list but not at the source list - rightfully).

gwsw:
You mean that it works with nmake on Windows, right? Do you think there's any compatibility issue with older Windows versions? Although I'm not sure if I actually care if we require Windows 11 to build.

Do you want to make the makefile change or do you want me to do it?

avih:
> ou mean that it works with nmake on Windows, right?

No. I don't care about the msvc file currently - it already has just one list.

I've been talking only about the mingw file the whole time, because it has two lists.

You can do it, but if you won't, I'll PR it at some stage. Not now.

gwsw:
Ok, cool. I tried it on nmake and apparently both patsubst and the $(SRC:.c=.o) syntax works there too. So I'll probably change both Makefile.wng and Makefile.wnm; the former to use one list instead of two, and the latter to use a list instead of *.c.

avih:
The current nmake file works with msvc6, so if you don't have it to test (I think I linked to a repo which has it not long ago), show me the patch before you push it and I can test it with msvc6 too, assuming you want to keep supporting it. It's from around the same time as borland 5.5, late 90's.

gwsw:
Thanks. I've made the makefile changes in PR #806.

PR #806, opened 2026-07-13
Create single list of source files in Windows makefiles.

gwsw:
Improve file handling in Makefile.wnm and Makefile.wng:

Makefile.wng previously contained two lists, LESS_SRC for source files and OBJ for object files. Unify these so that only LESS_SRC explicitly lists files. OBJ is derived from LESS_SRC using patsubst.

Makefile.wnm previously contained one list, OBJ of object files. Where the list of sources was needed, it just used "*.c". Change to make SRC a list of sources, and OBJ is derived from SRC using patsubst.

### DIFF
--- a/Makefile.wng
+++ b/Makefile.wng
@@ -109,36 +109,25 @@ LIBS += ${XLDLIBS}
 .c.o:
 	${CC} -c -I. ${CFLAGS} $<
 
-LESS_SRC = brac.c ch.c charset.c cmdbuf.c command.c		\
-           cvt.c decode.c edit.c evar.c filename.c forwback.c 		\
-           ifile.c input.c jump.c line.c linenum.c lmsg.c 		\
-           lsystem.c main.c mark.c optfunc.c option.c 		\
-	   opttbl.c os.c output.c pattern.c position.c 		\
-	   prompt.c          screen.c scrsize.c search.c 	\
-	   signal.c tags.c ttyin.c version.c xbuf.c
+LESS_SRC = brac.c ch.c charset.c cmdbuf.c command.c \
+           cvt.c decode.c edit.c evar.c filename.c forwback.c \
+           ifile.c input.c jump.c line.c linenum.c lmsg.c \
+           lsystem.c main.c mark.c optfunc.c option.c \
+           opttbl.c os.c output.c pattern.c position.c \
+           prompt.c screen.c search.c \
+           signal.c tags.c ttyin.c version.c xbuf.c
+
 ifeq  (${REGEX_PACKAGE},regcomp-local)
 LESS_SRC += regexp.c
 endif
+
 ifneq (${SECURE_COMPILE},1)
 LESS_SRC += lesskey_parse.c
-endif
-
-OBJ = \
-	main.o screen.o brac.o ch.o charset.o cmdbuf.o \
-	command.o cvt.o decode.o edit.o evar.o filename.o forwback.o \
-	help.o ifile.o input.o jump.o line.o linenum.o lmsg.o \
-	lsystem.o mark.o optfunc.o option.o opttbl.o os.o \
-	output.o pattern.o position.o prompt.o search.o signal.o \
-	tags.o ttyin.o version.o xbuf.o
-ifeq  (${REGEX_PACKAGE},regcomp-local)
-OBJ += regexp.o
-endif
-ifneq (${SECURE_COMPILE},1)
-OBJ += lesskey_parse.o
 else
 MINGW_DEFINES += -DSECURE_COMPILE=1
 endif
 
+OBJ = ${patsubst %.c,%.o,$(LESS_SRC)} help.o
 
 all: less.exe lessecho.exe
 
--- a/Makefile.wng
+++ b/Makefile.wng
@@ -25,8 +25,8 @@
 # REGEX_PACKAGE == gnu  (might not be available)
 #   This choice selects the GNU implementation, if provided by MingW.
 #
-# By default, the files help.c and funcs.h are generated using few posix
-# utilities (grep, sed, etc). On Windows, if the tools are missing, add:
+# By default, the files help.c, funcs.h and lessmsg.inc are generated using few
+# posix utilities (grep, sed, etc). On Windows, if the tools are missing, add:
 #   WINGEN=1
 # to compile a native C program which will be used instead of the posix tools.
 # NOTE: to cross compile with WINGEN=1, e.g. to arm, first run with native CC:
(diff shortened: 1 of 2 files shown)


avih:
The nmake change doesn't work for me in msvc2015 or msvc6. I don't have later msvc. It errors in both with the same error:
```
Makefile.wnm(50) : fatal error U1000: syntax error : ')' missing in macro invocation
Stop.
```

But this worked in both msvc2015 and msvc6:
```diff
diff --git a/Makefile.wnm b/Makefile.wnm
index 48f4b28..39839f7 100644
--- a/Makefile.wnm
+++ b/Makefile.wnm
@@ -47,7 +47,7 @@ SRC = $(SRC1) lesskey_parse.c
 CFLAGS = $(CFLAGS1)
 !ENDIF
 
-OBJ = $(patsubst %.c,%.obj,$(SRC)) help.obj
+OBJ = $(SRC:.c=.obj) help.obj
 
 # This rule allows us to supply the necessary -D options
 # in addition to whatever the user asks for.
```

Questions:
- At Makefile.wnm, why was the obj list var changed from OBJ1 to OBJ?
- At Makefile.wnm, why was it changed to begin with from OBJ to OBJ1 in a5548df2d48e0a91f97f7644cd4878aeeaffec40 ? also same question about changing CFLAGS to CFLAGS1 .
- Also in a5548df2d48e0a91f97f7644cd4878aeeaffec40 , why was the `.c.obj::` rule moved at Makefile.wnm to after the obj list, but in Makefile.wng the rule `.c.o:` was kept at the same place (before the sources list) ?

I'm guessing the OBJ1, SRC1, CFLAGS1 are because nmake doesn't support `+=` ? (does it really not?)

FWIW, `X = $(X) Y` seems to work as a replacement for `X += Y` in both msvc6/2015, though I don't know whether such self-reference is legal (but it is given as an example [here](https://stackoverflow.com/a/9848280/3339341), and menitions that the nmake expansion strategy is different than AT&T make, but I didn't find other sources for this):
```diff
diff --git a/Makefile.wnm b/Makefile.wnm
index 48f4b28..753139e 100644
--- a/Makefile.wnm
+++ b/Makefile.wnm
@@ -18,7 +18,7 @@
 CC = cl
 
 # Normal flags
-CFLAGS1 = /nologo /W3 /EHsc /O2 /I "." /D "WIN32" /D "NDEBUG" /D "_CONSOLE" /c
+CFLAGS = /nologo /W3 /EHsc /O2 /I "." /D "WIN32" /D "NDEBUG" /D "_CONSOLE" /c
 LDFLAGS = /nologo /subsystem:console /incremental:no
 
 # Debugging flags
@@ -30,7 +30,7 @@ LIBS = user32.lib shell32.lib
 
 #### End of system configuration section. ####
 
-SRC1 = brac.c ch.c charset.c cmdbuf.c command.c \
+SRC =  brac.c ch.c charset.c cmdbuf.c command.c \
        cvt.c decode.c edit.c evar.c filename.c forwback.c \
        ifile.c input.c jump.c line.c linenum.c lmsg.c \
        lsystem.c main.c mark.c optfunc.c option.c \
@@ -40,14 +40,12 @@ SRC1 = brac.c ch.c charset.c cmdbuf.c command.c \
        regexp.c
 
 !IF "$(SECURE_COMPILE)" == "1"
-SRC = $(SRC1)
-CFLAGS = $(CFLAGS1) /D "SECURE_COMPILE=1"
+CFLAGS = $(CFLAGS) /D "SECURE_COMPILE=1"
 !ELSE
-SRC = $(SRC1) lesskey_parse.c
-CFLAGS = $(CFLAGS1)
+SRC = $(SRC) lesskey_parse.c
 !ENDIF
 
-OBJ = $(patsubst %.c,%.obj,$(SRC)) help.obj
+OBJ = $(SRC:.c=.obj) help.obj
 
 # This rule allows us to supply the necessary -D options
 # in addition to whatever the user asks for.
```

EDIT: it is mentioned at the [official nmake docs](https://learn.microsoft.com/en-us/cpp/build/reference/special-nmake-macros?view=msvc-170) at the "Environment-variable macros" section with this example (after it states that env vars are inherited as nmake macros):
```make
PATH=$(PATH);\nonesuch
```
so the patch above seems to be legal in nmake, and those `1` can be removed.

Also, the comment for the "generated" rule in Makefile.wnm could mention `lessmsg.inc` too. Similarly at Makefile.wng (the comment about WINGEN=1).

gwsw:
Yes, nmake apparently does not support ```+=```. I also have a vague memory that some version of make didn't even allow reassigning a make variable, but that might not have been nmake (maybe Borland)? Your changes work for me, so I've updated the PR.

I assume mingw uses gnu make? I tested the mingw build using a cross compiler on Linux with the Linux gnu make. I don't know if it's necessary to also test a native mingw on Windows.

avih:
LGTM and the changes work for me on both msvc6 and msvc 2015. Thanks also for the lessmsg.inc comments.

Which leaves:

> ... in [a5548df](https://github.com/gwsw/less/commit/a5548df2d48e0a91f97f7644cd4878aeeaffec40) , why was the `.c.obj::` rule moved at Makefile.wnm to after the obj list, but in Makefile.wng the rule `.c.o:` was kept at the same place (before the sources list) ?

avih:
> I assume mingw uses gnu make?

Mingw is the toolchain (and binutils) and the windows headers. `make` is not part of mingw.

There can be gcc-based mingw - like w64devkit, and clang-based mingw - like llvm-mingw.

I think binutils are cross-compatible natively, i.e. you can use any version of binutils `strip` or `ar` etc, and it would work correctly depending on the input binary format. So effectively mingw is just the compiler (which creates windows PE binaries) and windows headers.

On linux if you have mingw cross-toolchain installed then you typically use the system's `make` if you need it.

On windows most mingw distributions also include gnu make and sometimes other utilities (cmake, etc).

So mingw is typically used with gnu make, but only due to circumstances.

However, regardless of that, the less Makefile.wng already depends on gnu make with `ifeq` etc, and has been from day 1 of this file, so in less it does require gnu make.

Also, I did test the Makefile.wng changes too - which do work. Thanks.

gwsw:
I don't remember exactly why I made that change. I would guess that I was thinking (whether based on evidence or a guess I don't know) that the use of $(CFLAGS) in the rule should follow its definition. Since a5548df moved the definition of CFLAGS, I moved the .c.obj rule to follow it.

avih:
> Since [a5548df](https://github.com/gwsw/less/commit/a5548df2d48e0a91f97f7644cd4878aeeaffec40) moved the definition of CFLAGS, I moved the .c.obj rule to follow it.

Thanks.

PR #818, opened 2026-08-17
Add opt-in temporary-file spooling for standard input

CouscousPie:
Many programs use less as their pager by piping generated output to standard
input. In that mode, less identifies the input as `-`, so the `v` command has
no editor-readable filename and reports "Cannot edit standard input".

This creates a downstream integration problem. Pager users get normal less
navigation and search, but tools which rely on `v` cannot offer an editor
handoff unless they implement their own temporary-file handling, change how
they invoke the pager, or disable that workflow. Those workarounds duplicate
temporary-file lifecycle, cleanup, quoting, and security handling outside less.

Add the opt-in `--spool-stdin` option. When less is invoked with
implicit standard input, it copies that input to a secure temporary file before
displaying it and opens the file normally. The existing `v` command therefore
works without a separate editor code path. The temporary file is removed when
less exits.

This is deliberately opt-in because it changes stdin handling: less must read
the complete input stream before it can display it, and it consumes temporary
disk space proportional to the input.

The option can also be supplied through `LESS`, for example:

```sh
some-command | LESS=--spool-stdin less
```

Validation:

- Prepared the git checkout with `make -f Makefile.aut all`.
- Full build succeeded.
- `make check` passed: 19 tests, 0 errors.
- A PTY smoke test piped `alpha\nbeta\n` into less, invoked `v`, verified that
  the editor received a `/tmp/less-stdin-*` path containing that exact input,
  and verified that the path was removed after less exited.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -293,7 +293,7 @@ AC_MSG_CHECKING(for ANSI function prototypes)
 AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[]], [[int f(int a) { return a; }]])],[AC_MSG_RESULT(yes); AC_DEFINE(HAVE_ANSI_PROTOS)],[AC_MSG_RESULT(no)])
 
 # Checks for library functions.
-AC_CHECK_FUNCS([fchmod fsync nanosleep poll popen realpath _setjmp sigprocmask sigsetmask snprintf stat strsignal system ttyname usleep])
+AC_CHECK_FUNCS([fchmod fsync mkstemp nanosleep poll popen realpath _setjmp sigprocmask sigsetmask snprintf stat strsignal system ttyname usleep])
 AC_CHECK_DECL(sigsetjmp, [AC_DEFINE(HAVE_SIGSETJMP)], [], [#include <setjmp.h>])
 
 # AC_CHECK_FUNCS may not work for inline functions, so test these separately.
--- a/less.hlp
+++ b/less.hlp
@@ -286,6 +286,8 @@
                   Set default options for every search.
                 --show-preproc-errors
                   Display a message if preprocessor exits with an error status.
+                --spool-stdin
+                  Spool standard input to a temporary file before viewing it.
                 --past-eof
                   Scrolling commands continue past end of file.
                 --proc-backspace
--- a/less.nro.VER
+++ b/less.nro.VER
@@ -2157,6 +2157,9 @@ If a LESSOPEN preprocessor produces data,
 then exits with a non-zero exit code,
 .B less
 will display a warning.
+.IP "\fB\-\-spool\-stdin\fP"
+When viewing standard input, copy it to a temporary file before displaying it.
+The temporary file is removed when \fBless\fP exits.
 .IP "\fB\-\-status\-col\-width=\fP\fIn\fP"
 Sets the width of the status column when the
 .B "\-J"
--- a/lessmsg
+++ b/lessmsg
@@ -331,6 +331,8 @@ Mouse_features Mouse features:\40
 Header_lines Header lines:\40
 Warn_when_editing_a_file_opened_via_LESSOPEN Warn when editing a file opened via LESSOPEN
 Dont_warn_when_editing_a_file_opened_via_LESSOPEN Don't warn when editing a file opened via LESSOPEN
+Dont_spool_standard_input Don't spool standard input to a temporary file
+Spool_standard_input Spool standard input to a temporary file
 interrupt_character interrupt character:\40
 Search_match_shift Search match shift:\40
 Autosave_actions Autosave actions:\40
--- a/main.c
+++ b/main.c
@@ -37,6 +37,7 @@ public POSITION start_attnpos = NULL_POSITION;
 public POSITION end_attnpos = NULL_POSITION;
 public int      wscroll;
 static constant char *progname;
+static char *stdin_spoolfile;
 public lbool    quitting = FALSE;
 public lbool    dohelp = FALSE;
 public char *   init_header = NULL;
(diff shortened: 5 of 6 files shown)


gwsw:
I have a couple of questions about this.

First, in what situations is it useful to edit a temp file that will be removed when less exits? I can't see a use case where a user would want to edit a file that is soon going to disappear. It also seems likely to cause confusion if a user enters the ```v``` command and thinks that they are editing a permanent file and then can't find it after exiting less.

Also, if a user does want this functionality, they could set their PAGER to a simple shell script like
```
if [ $# -gt 0 ]; then less $*; else T=$(mktemp); cat >$T; less $T; rm $T; fi
```
Wouldn't this have the same effect without requiring a change to less?

gwsw:
Also, this is a minor point and is moot if this feature is not added, but I don't think we should use the term "spool" for the option name or in documentation. It's a fairly obsolete computing term and probably wouldn't be understood by most users, or even most younger computer professionals. I don't think I've heard it used since the 1970s.

CouscousPie:
I have to admit: you are absolutely right. A wrapper works. I rediscovered I even had such a wrapper written already. So, this draft's only remaining purpose would be a convenience layer. Is there a specific reason or principle why it should not be in less?

gwsw:
Well, it seems much less likely for there to be a bug or unexpected interaction using a one-line shell script vs. the 100 or so lines that this PR would add to less. The more features that are added to less, the more likely it is that there will be some interaction with other features or other operating environments that cause some unexpected behavior or bug. For example, the PR uses ```atexit()``` and ```unlink()```, neither of which are currently used anywhere else in less. Are these functions available in all environments where less is currently supported? MS-DOS? OS/9? Borland or djgpp or mingw compilers? etc.   And even the build doesn't break, what runtime issues might there be? Are there environments where 8092 is too large for a stack variable? Is ```--help``` the only case where we should bypass the spooling? What about ```--version```? Thinking of and testing all such interactions would be difficult and any issues may not be discovered for a long time. 

I think the question should be, is there any reason to add the feature to less if the desired behavior can be achieved fairly simply without doing so?

CouscousPie:
To me personally, it seems preferable to have desired features inside a tool instead of writing wrappers (not only, but also because fiddling with PATH and environments becomes obsolete and the tool behaves the exact same anywhere) and I feel like editing stdin is a pretty basic feature, especially as this is the most effective way to get information from the pager to somewhere else without using a mouse and selecting it via the terminal emulator as an escape hatch. In this particular case it might be generally just a simple script to add it on, but when the problem appears downstream like for me while using bat it can become difficult to even identify the underlying issue. Adding such a flag might help making the issue easier discoverable and it also might signal to maintainers downstream that it is a desirable feature and just for example hint at them implementing documentation in this regard.

Anyway, if you think it is of too little value, I'll just close the drafted PR.

gwsw:
Well, different viewpoints I guess. To me, using an editor just to move some data from less into another file or program seems a rather roundabout approach. I would either use the ```s``` command if I want to save the whole file, or the ```|``` command if I want to save only part of the file, or pipe it to a program.

CouscousPie:
Yes, it looks like it. Maybe a pager is just not for me? Because I like to view things in pagers, but what then often happens is that only when I am already looking at the actual page I realize there is something, maybe a git ref, an IP address or a command from a man page, that I want to have in my clipboard ready for the next command or store it somewhere. Currently vanilla less is a predetermined end to any pipe, which is fine in many cases, but an unnecessary limitation in all the others.

avih:
I think you still didn't answer the main question: what is it good for?

I.e. which reasonable use case requires editing a file from stdin, especially if it's deleted once less exits? how is that useful?

CouscousPie:
For example compensating for bad working memory and allowing to continue to work in one continuous flow with any data. I think my last comment made this clear.

avih:
> For example compensating for bad working memory and allowing to continue to work in one continuous flow with any data. I think my last comment made this clear.

I don't understand this. You mean this?
> only when I am already looking at the actual page I realize there is something, maybe a git ref, an IP address or a command from a man page, that I want to have in my clipboard ready for the next command or store it somewhere.

If yes, how does editing the file helps you with that? the file is deleted anyway when less exits, so how does editing it lets you keep the data if it's deleted?

Also, you could just copy to clipboard from less itself, by selecting with the mouse.

Or you could dump some part of the content (or the whole content) to some file from within less if you want to keep some part of it.

I still don't see how editing a file which is then deleted helps in any way at all.

avih:
> Or you could dump some part of the content (or the whole content) to some file from within less if you want to keep some part of it.

For reference, the `s` command lets you save piped input to a file of your choice.

Alternatively, the `|` command lets you pipe the current screen or some other parts of the content to a shell command, for instance:
```
|. cat > myfile
```
will save the current screen to the file `myfile`.

CouscousPie:
Let me try to make the use case more concrete and let me be upfront, that I let an AI restructure and distill my draft. So, the surface is mostly AI generated, the content is not.

The thing `v` gives you that `s` and `|` don't is interactive selection of a piece of the screen you didn't know you'd want in advance. `s` writes the whole stream to a file. `|` runs a shell command over a screen region, which is fine when you already have the pattern in your head. Neither of them lets you point at a specific bit of what's on the screen and grab it without describing it first.

Concrete example. `ip address | less`, screen shows:

```
    inet6 fe80::3e7c:3fff:fea1:b2c3/64 scope link noprefixroute
```

I want `fe80::3e7c:3fff:fea1:b2c3`.

With `|` today, I'd type something like:

```
|.  grep -o 'fe80[^[:space:]]*'
```

To receive an exact match. The more ambiguous the desired string, the more I have to type which is vastly decreasing the actual value of doing so. Even in this unambiguous case, the regex is almost as long as the string I want, arguably even more complicated, difficult to type (note that not everybody has a US keyboard layout) and error prone.

With `v` on stdin, in vim the limitless options of continuing to work with that piece of information start with one of two options that immediately come to mind:

1. searching with `/fe80<CR>`
2. selecting with visual navigation

Then simply hitting for example
```
viWy
```
The address is then in the unnamed register. From the exact same session I can paste it into a file (`:e ~/notes`, then `p`), use it as a command argument (`:!ssh -6 <C-r>"`), or send it to a clipboard tool. It also scales indefinitely (I can `:!echo <C-r>" | tee -a **` to append it to all files of current directory with a single copy/paste). I never had to know the exact string, and I never wrote any regex. I *selected* data without previous knowledge rather than *matching* it.

**On "the file gets deleted, how does that help":** the file isn't what I'm trying to keep. It's a scratch surface. I want information that is walled off inside it. I want a selection ending up somewhere that outlives less: an editor register, another open buffer, or a shell command line I'm about to run. Once it's there, the temp file can go.

**On the wrapper:** you're right, a wrapper works for me. I rediscovered mine, as I said. What the wrapper doesn't do:

- help anyone using a tool that calls less directly instead of via `$PAGER`. `git`, `man`, `systemctl`, `journalctl`, `bat` and others do this in one config or another.
- help anyone hitting `Cannot edit standard input` for the first time. That message reads as a hard limit of less, with nothing pointing them at the wrapper.
- give downstream tool maintainers something concrete to link to in their own docs when explaining how to hand off from their pager to an editor.

A flag in less would turn this from folklore into documented behavior.

**On mouse-selecting from the terminal:** that's a terminal-emulator feature, not a less feature. It needs a graphical terminal with a working pointer, so it doesn't help on a Linux virtual console (`Ctrl+Alt+F<n>`), over a serial console, or in a tmux or screen session that doesn't have mouse mode on. `v` is the keyboard route from pager to editor, and right now it doesn't work on stdin.

This is not only about personal preference, it is also about accessibility. I already mentioned neurological diversity and now with physical pointing devices being brought up, I feel like I have to also mention that not everybody is physically able to move a mouse, for permanent or temporary reasons that might not yet be obvious to you. Please also forgive me for having to point out the irony of maintainers of *the* terminal pager to route me towards usage of a device unavailable in and not native to actual terminals.

If the position is still that this belongs entirely outside less, a much smaller change would still help: extend the `Cannot edit standard input` error and the man page entry for `v` to mention the wrapper recipe. That doesn't grow less's surface area, but it at least tells people hitting the limit that there's a way out.

avih:
When I said that `s` can save the input, I meant that less already has everything you want to open it in an editor without adding new code in less.

Make some binding of your own which runs `sSOMENAME; !vim SOMENAME; !rm SOMENAME` to save it into a file, open it in vim (or whatever editor you want), and delete it once you exit vim.

This does not require new features in less even if you did not have your one-line wrapper which does that too.

This is untested, but you can make it work, and use mktemp too for the name, or some other commands which less already supports, without adding new code.

CouscousPie:
Thank you for your responsiveness and time.

PR #821, opened 2026-09-02
Git repository improvements (1/4)

ralish:
This PR contains a small set of improvements primarily to the Git repository itself. In short it:
- Tidies up the `.gitignore` and adds some additional exclusions to capture additional build outputs
- Adds a `.gitattributes` to ensure shell scripts are always checked-out with LF EOL sequences
- Trims EOL whitespace and multiple EoF newlines across the entire repository
- Adds a `.git-blame-ignore-revs` to ignore the whitespace clean-up in `git blame`
- Adds a `.mailmap` to consolidate multiple author identities in the commit history

Further details are available in each commit description. None of this functionally changes the less source code itself, but it does make working in the Git repository a little nicer.


ralish:
I wanted this to be a stacked PR but apparently they don't work across repositories, so I'd suggest reviewing and merging them in sequence once you're comfortable with them. That way the diff of the subsequent PR will be limited to just its changes. If you want changes made I'll make the edits and flow them through the remaining PRs.

